Issue #15, opened 2026-04-30
Upstream: @wordpress/ui overlay components blocked on WP core privateApis allowlist


dabowman:
`Notice`, `Tooltip`, `Popover`, `Dialog`, `AlertDialog`, `Drawer`, `IconButton`, form `Select`/`Autocomplete` from `@wordpress/ui` all transitively pull `@wordpress/theme`, which isn't in WordPress 6.9 core's `wp.privateApis` allowlist (`CORE_MODULES_USING_PRIVATE_APIS`).

The Gutenberg plugin overrides `wp-private-apis` with one that includes `@wordpress/theme` / `@wordpress/ui` / `@wordpress/dataviews` / etc., so the shell works locally because Gutenberg is in `.wp-env.json`. Production must declare `Requires Plugins: gutenberg` (planned for M5).

Currently using `@wordpress/components` fallbacks where available. Re-evaluate when WP core allowlists `@wordpress/theme` natively.

Source: pre-M1 inbox. Logged in `docs/feedback.md` + `CLAUDE.md` Key rules.

dabowman:
Mitigation shipped in M5: `Requires Plugins: gutenberg` plugin header + admin-notice fallback when Gutenberg is missing (commit 0816737). Keeping open as a passive tracker — when WordPress core adds `@wordpress/theme` / `@wordpress/ui` / `@wordpress/dataviews` to the `CORE_MODULES_USING_PRIVATE_APIS` allowlist, we can drop the Gutenberg dependency and the admin notice. No action on our side until then.

dabowman:
Closing — resolved upstream. WordPress 7.0 ships the `CORE_MODULES_USING_PRIVATE_APIS` allowlist including `@wordpress/theme` / `@wordpress/ui` / `@wordpress/dataviews` (and bundles `@wordpress/theme`) in core, which is exactly what this tracker was waiting for.

PR #249 acted on it: the `Requires Plugins: gutenberg` header is gone, and `wp_admin_workspaces_dependencies_met()` version-gates the runtime — WP ≥ 7.0 needs nothing extra; WP 6.7–6.9 falls back to detecting the Gutenberg plugin (admin notice + hijack stand-down when unmet). Verified live on WP 7.0 with Gutenberg deactivated.

The `@wordpress/components` fallbacks for WPDS 0.12 gaps are a separate concern (per-component, documented in CLAUDE.md) and not blocked on this.

Issue #14, opened 2026-04-30
DX: lint rule for iconMap keys + recurring null/state-refresh patterns


dabowman:
Recurring code-review patterns keep showing up:
- Missing null guards on entity records.
- Stale state after mutations (forgetting to refetch).
- Icon name string mismatches against `src/runtime/config/iconMap.js`.

Consider a lint rule or pre-commit check for `iconMap.js` keys (the surface is small and string-typed). Null-guard patterns and state-refresh patterns are harder to enforce mechanically — at minimum, codify them in `CLAUDE.md` so the rules are explicit when reviewing.

Source: pre-M1 inbox. Logged in `docs/feedback.md`.

dabowman:
Fixed in c88fd5897932f8d8a54eed6f8a4e4723f66f32d2. Dev-mode warn-on-miss in resolveIcon + null-guard / state-refresh / icon-map patterns codified in CLAUDE.md.

Issue #13, opened 2026-04-30
EditorApp: seed placeholder block markup on new-post auto-draft


dabowman:
The new-post flow in `src/apps/EditorApp.js` doesn't seed placeholder block markup, so the REST API rejects with `Content, title, and excerpt are empty`.

`SimpleEditorApp` already seeds:
\`\`\`html
<p></p>
\`\`\`

Mirror the fix into `EditorApp.js` next time it's touched.

Source: pre-M1 inbox + M1 review. Logged in `docs/feedback.md`.

dabowman:
Fixed in 2f571eeb4391f0c874820207712b577cf12d69f4. Auto-draft now seeds the same placeholder paragraph SimpleEditorApp uses.

Issue #12, opened 2026-04-30
Extract shared DataViews-app base from PostsApp / UsersApp / CommentsApp


dabowman:
`src/apps/PostsApp.js`, `src/apps/UsersApp.js`, `src/apps/CommentsApp.js` follow the same DataViews pattern: `useEntityRecords` → fields → actions → render. Each has domain-specific logic (custom fields, actions, filtering), but the wiring is repeated.

Acceptable for v1 idiom. If more DataViews-backed apps land in v2 (e.g., taxonomies, forms, custom CPTs), extract a shared base component or hook to keep the surface consistent.

Source: M4 review. Logged in `docs/feedback.md`.

dabowman:
Punting to v2 per the issue's own framing — refactor when a 4th+ DataViews-backed app lands. Three apps don't justify the abstraction yet.

PR #40, opened 2026-05-14
feat(c5): lazy app registration + per-app webpack chunks

dabowman:
## Summary

Track D — Lazy App Registration. Registry's `register()` accepts both eager `{ Component }` and lazy `{ load: () => import(...) }` shapes; mount path awaits the load on first match through `registry.resolveComponent(id)` → `React.lazy()` wrapped in `<Suspense>` + `<AppErrorBoundary>`. Each lazy app emits its own webpack chunk (`build/app-<id>.js`) via `webpackChunkName` magic comment.

Modeled on CIAB's route+content split. Different mechanism (webpack dynamic `import()` vs CIAB's `wp_register_script_module`), same outcome — only pay for what mounts.

## Bundle impact

Captured 2026-05-14, `npm run build` production.

| Asset | Pre-D | Post-D | Δ |
|---|---:|---:|---:|
| `index.js` (boot) | 2,164,072 B (2.06 MiB) | 213,691 B (209 KiB) | **−90.1 %** |
| `index.css` | 28,735 B | 20,860 B | −27 % |
| Entrypoint (cold-mount JS+CSS) | 2.12 MiB | 250 KiB | **−88 %** |
| Vendors chunk `245.js` (lazy) | — | 1.75 MiB | new — loads on first DataViews-backed mount |
| Lazy `app-*.js` chunks | — | 25 chunks, 973 B – 17 KiB | new |

Auto-extracted vendors chunk is still big but lazy. Splitting it further is Track E candidate.

## Migration shape

- Always-eager (5): `core:navigation`, `core:site-hub`, `core:toolbar-actions`, `core:notices-banner`, `core:notices-snackbar` — mounted in every shell.
- Lazy (25): everything else, registered as `{ load: () => import(/* webpackChunkName: "app-<id>" */ '../../apps/<id>') }`.

## Changes

- `src/runtime/registry/createRegistry.js` — accepts `load`, rejects both-set, adds `resolveComponent(id)` w/ per-id Promise cache; engines stay eager-only.
- `src/runtime/regions/mountApp.js` — `React.lazy` wrapper memoized in per-id `lazyAppCache`; `<Suspense>` + `<AppErrorBoundary>` + retry that clears the cache.
- `src/runtime/registry/builtins.js` — 25 of 30 apps flipped to dynamic `import()` w/ magic-comment chunk names.
- `src/index.css` — minimal kernel-CSS for `.wp-admin-shell-app-loading` (reserves space) + `.wp-admin-shell-app-error` (structural padding only — DS-neutral).
- `tests/runtime/registry-lazy-app.test.mjs` — new, 20 assertions: both shapes, both-set rejection, cache identity, error paths, engine-lazy rejection, hydration.
- `package.json` — adds new test to `test:runtime` chain.
- `CLAUDE.md` + `docs/v1-perf-baseline.md` — status block, file-tree update, bundle-size table.

## Test plan

- [x] `npm run test:runtime` — 350/350 (was 330; new file adds 20)
- [x] `npm run test:schema` — 79/79
- [x] `npm run test:parity` — 4/4
- [x] `npm run test:engines` — 77/77
- [x] `npm run lint:ts` — clean
- [x] `npm run build` — 25 `app-*.js` chunks emit; `index.js` 213,691 B; vendors chunk 1.75 MiB
- [ ] **Manual browser smoke (reviewer):** open shell, confirm Network panel shows boot `index.js` + only the routed app's chunk loads on first paint; navigate to another route, confirm a fresh chunk request fires; force a chunk 404 (e.g. break the chunkhash) and confirm the inline `AppLoadError` retry button renders.


src/runtime/regions/mountApp.js
```diff
@@ -36,6 +50,115 @@ function warnDsMismatch( engineId, appId, engineDs, appDs ) {
...
+
+function retryLazyApp( appId ) {
+	// Drop the cached React.lazy wrapper so the next render rebuilds
+	// it. The registry's resolveComponent cache is also wiped via the
+	// load thunk being re-invoked. (Note: the registry itself does not
+	// expose a public invalidate — the most common cause of a load
+	// failure is a chunk fetch error, which webpack's chunk-loading
+	// runtime already retries internally. The retry button mostly
+	// re-tries any lazy() that errored during render.)
+	lazyAppCache.delete( appId );
```
dabowman, line 206: **Retry is a no-op for chunk-load failures.** `lazyAppCache.delete(appId)` rebuilds the `React.lazy` wrapper, but `getLazyComponent` (line 60-73) calls `registry.resolveComponent(id)` which returns the *cached rejected promise* from `loadCache` in `createRegistry.js:137-139`. The new `lazy()` wrapper resolves to the same rejection synchronously — error boundary fires again, identical state.

The inline comment downplays this with "webpack's chunk-loading runtime already retries internally" — that's not accurate. Webpack's `__webpack_require__.f.j` only retries on a few transient signals (`webpackChunkLoadTimeout`); 404s / network errors do not auto-retry. The `import()` promise rejects once and stays rejected for the lifetime of that registry instance.

Fix options:
- Expose `registry.invalidateComponent(id)` that deletes from `loadCache` (and clears `source.Component` if hydrated) and call from `retryLazyApp`. Then retry actually re-fires `load()`.
- Or: drop the retry button, document recovery as "reload the shell", and stop the comment from claiming retry helps.

Also worth a runtime test: assert `load()` call-count goes from 1 → 2 after `invalidate + resolveComponent` once the API exists. The current `tests/runtime/registry-lazy-app.test.mjs` covers neither retry nor cache-invalidation.

src/runtime/regions/mountApp.js
```diff
@@ -102,12 +240,19 @@ export function MountedApp( { appRef, regionId, segments, fallback = null } ) {
...
 			>
-				<Component
-					app={ appInstance }
-					config={ mergedConfig }
-					regionId={ regionId }
-					segments={ segments || [] }
-				/>
+				<AppErrorBoundary
+					appId={ appInstance.source }
+					onRetry={ retryLazyApp }
```
dabowman: **`AppErrorBoundary` now wraps eager chrome apps too**, swallowing any render-time error and painting `"Failed to load app \"core:navigation\""`. Pre-C5, a render crash in `core:navigation` would surface to whatever top-level boundary the host had (or crash the tree visibly in dev).

The copy is wrong for that case — eager apps don't "fail to load", they fail to render. Either:
- Mount the boundary only when the source is lazy (`if (!sourceDef.Component)`), keeping the pre-C5 surface for chrome apps; or
- Make the message branch on whether the error is a chunk-load failure (`error.name === 'ChunkLoadError'` for webpack 5) vs a generic render error.

Right now a typo in NavigationApp during dev silently turns into a vague "failed to load" panel.

src/runtime/regions/mountApp.js
```diff
@@ -36,6 +50,115 @@ function warnDsMismatch( engineId, appId, engineDs, appDs ) {
...
+	);
+}
+
+function AppLoadError( { appId, error, onRetry } ) {
+	const message = error?.message || String( error );
+	return (
+		<div className="wp-admin-shell-app-error" role="alert">
+			<p>
+				{ __( 'Failed to load app', 'wp-admin-shell' ) }
+				{ appId ? ` "${ appId }"` : null }.
```
dabowman: i18n nit: string is built by concatenating `__('Failed to load app')` + literal `" \""` + `appId` + `"."`. Translators can't reorder around the appId. Use `sprintf( __( 'Failed to load app "%s".', 'wp-admin-shell' ), appId )` from `@wordpress/i18n`. (When appId is missing, fall back to the bare `__( 'Failed to load app.' )`.)

src/runtime/registry/createRegistry.js
```diff
@@ -69,5 +108,70 @@ export function createRegistry() {
...
+				if ( typeof Component !== 'function' ) {
+					throw new Error(
+						`createRegistry: load() for "${ id }" did not resolve to a React component`
+					);
+				}
+				// Hydrate the registry entry so synchronous consumers
+				// (the legacy `Component` read in `mountApp`) start
+				// hitting the eager path after first resolution. The
+				// Promise still owns identity for Suspense.
+				source.Component = Component;
```
dabowman: `source.Component = Component;` mutates a registered descriptor from inside a query method. It's load-bearing — `mountApp.js:221-222` does `sourceDef.Component || getLazyComponent(...)` and relies on this hydration to skip the lazy path on the second mount.

Works fine, but it couples register-time invariants (the `Component` XOR `load` rejection at line 68) to the read path: after `resolveComponent`, the descriptor now violates that invariant (has both). If anything ever re-validates a stored source, it'll trip.

Alternative that keeps invariants clean: keep `loadCache` authoritative (already there), and have `mountApp` always read through `resolveComponent` / a `getResolvedComponent(id)` sync accessor that returns the cached value or null. Same fast-path, no descriptor mutation. Not a blocker; flagging because the test for it (test #12) treats the mutation as the contract.

dabowman (commented):
Track-D bundle wins are real (90 % off boot bundle, separate vendors chunk). Four issues to look at before merge — the first is the only one I'd call a bug; rest are nits.

1. **`retryLazyApp` is effectively a no-op for chunk-load failures** — see inline on `mountApp.js`.
2. **`<AppErrorBoundary>` swallows render errors from eager chrome apps** with misleading copy.
3. i18n concatenation in `AppLoadError`.
4. `source.Component = Component` mutation inside a query method.

Smoke item from the PR description ("force a chunk 404 → retry button renders") will pass even with #1 broken — the button paints, just doesn't recover. Worth either adding a registry-side invalidation or adjusting the comment to stop claiming retry works.

dabowman:
Pushed `24072ae` addressing all four review items.

**1. Retry no-op (the bug):** added `registry.invalidateComponent(id)` to `createRegistry.js`. Drops both `loadCache` (Promise) and `resolvedCache` (sync) for lazy ids; no-op for eager (their Component is part of the boot bundle, nothing to re-fetch). `retryLazyApp` in `mountApp.js` now calls it before clearing `lazyAppCache`. Two new tests confirm `load()` call-count goes 1 → 2 after invalidate, and a once-failed-then-recovers thunk succeeds on the retry.

**2. Boundary swallowing eager render errors:** mount path now branches on `sourceDef.Component`:

- **Eager** (`{ Component }` at register time): renders direct, no boundary, no Suspense. Render crashes in `core:navigation` propagate as they did pre-C5.
- **Lazy** (`{ load }`): still wrapped, but `AppErrorBoundary.isChunkLoadError()` filters by error type (`ChunkLoadError`, `Loading chunk …`, `createRegistry: load() …`). Non-load errors re-throw from `render()` so a post-resolution crash inside a lazy app surfaces with its real stack instead of our "Failed to load" copy.

**3. i18n concat:** replaced with `sprintf( __( 'Failed to load app "%s".' ), appId )` + a bare `__( 'Failed to load app.' )` fallback when `appId` is missing.

**4. Descriptor mutation:** dropped. `resolveComponent` now populates a separate `resolvedCache` map keyed by id; `getResolvedComponent(id)` is the new sync accessor (apps-only, returns null for unresolved lazy / unknown / engines). The stored descriptor's `Component XOR load` invariant holds for the life of the registration. Test #12 rewritten to assert this (`got?.Component === undefined && typeof got?.load === 'function'` after resolve).

Runtime suite 350 → 365 assertions. Boot bundle 213,691 → 214,491 B (+800 B for the error filter + sprintf import).

Smoke item still pending reviewer (need wp-env up).

src/runtime/regions/mountApp.js
dabowman: Fixed in 24072ae.  now calls  before clearing the per-id React.lazy wrapper, and createRegistry exposes  that drops both  (Promise) and  (sync). Test #14 covers the 'once-rejected then recovers' scenario explicitly — second resolve hits a different code path after invalidate, attempt count goes 1 → 2.

src/runtime/regions/mountApp.js
```diff
@@ -36,6 +50,166 @@ function warnDsMismatch( engineId, appId, engineDs, appDs ) {
...
+						appId={ this.props.appId }
+						error={ error }
+						onRetry={ this.retry }
+					/>
+				);
+			}
+			// Not a load failure — propagate so the upstream boundary
+			// (or React's dev overlay) surfaces the real error. Clear
+			// state first so we don't loop on subsequent renders.
+			throw error;
```
dabowman, line 193: Comment lies — no state clear before `throw error`. Throw during render unmounts this boundary, so loop is impossible regardless. Suggest dropping the second sentence or actually clearing (`this.setState({error: null})` would itself loop, so just drop the sentence).

src/runtime/registry/createRegistry.js
```diff
@@ -69,5 +126,133 @@ export function createRegistry() {
...
+	 *
+	 * Distinguishing "not loaded yet" from "loaded and resolved to a
+	 * real component" lets the mount path skip the React.lazy wrapper
+	 * once a chunk is cached, avoiding an extra Suspense flash on
+	 * re-mounts.
+	 *
+	 * @param {string} id App source id.
+	 * @return {Function|null} Cached component reference, or null.
+	 */
+	function getResolvedComponent( id ) {
```
dabowman: `getResolvedComponent` is exported + tested but no consumer calls it. `mountApp.js:298` still branches on `sourceDef.Component` directly, lazy path goes through `getLazyComponent` → `resolveComponent` (Promise) → `React.lazy`. After first resolve, `React.lazy` on a settled Promise renders synchronously already, so the sync-cache short-circuit isn't load-bearing.

Either wire `getResolvedComponent` into the mount path (skip the React.lazy wrapper once cache is hot) or drop the API as YAGNI. Currently it's a tested-but-unused surface.

dabowman (commented):
Fixes look good — all four prior items addressed cleanly:

1. **Retry recovery wired through** `registry.invalidateComponent(id)` clearing both `loadCache` + `resolvedCache`. Tests #13 (`load()` count 1→2 after invalidate) + #14 (rejected-then-recovers thunk) lock the contract.
2. **Boundary scope tightened.** Eager apps render bare (no boundary, no Suspense). Lazy boundary's `isChunkLoadError` filter catches `ChunkLoadError` / `Loading chunk …` / `Loading CSS chunk …` / `createRegistry: load() …`; everything else re-throws so render crashes hit outer boundary with real stack.
3. **i18n** `sprintf( __( 'Failed to load app "%s".' ), appId )` + bare-string fallback. Translator comment present.
4. **Mutation gone.** Descriptor's `Component XOR load` invariant holds for life. Test #12 asserts post-resolve descriptor stays unmutated; new `getResolvedComponent` covers the sync-cache contract.

365 runtime assertions pass locally. Two minor nits inline — non-blocking.

dabowman:
Pushed `5bdbca2` for both nits.

**Boundary comment:** dropped the false "Clear state first" sentence — throw-during-render unmounts the boundary so there's no loop regardless.

**`getResolvedComponent` YAGNI:** dropped the API. Reasoning matches yours — `React.lazy` on a settled Promise already renders synchronously, so the sync-cache short-circuit wasn't load-bearing. `resolvedCache` Map, register-time + resolve-time writes, and the `getResolvedComponent` accessor all gone. `invalidateComponent` simplifies to a single `loadCache.delete`. Only `loadCache` (Promise identity per id) remains; the descriptor's `Component XOR load` invariant still holds for life.

Runtime suite 365 → 359 (dropped 6 sync-accessor assertions; kept descriptor-non-mutation, load-recovery, invalidate-on-unknown, eager-invalidate-noop). Boot bundle 214,491 → 214,289 B.

dabowman (commented):
Both nits resolved.

* **`getResolvedComponent` + `resolvedCache` deleted.** Right call — React.lazy memoizes resolved component on settled Promise so sync cache was redundant. `invalidateComponent` simplifies to single `loadCache.delete`. Eager-invalidate test (#15) updated to assert recovery via re-resolve.
* **Boundary throw comment trimmed.**

359 runtime assertions pass locally. Bundle holds at 209 KiB (−90 % vs pre-D). No further blockers from me.

dabowman:
LGTM. Implementation matches the plan; tests green in the worktree; bundle savings real and measured.

## Verified against plan

| Plan item | Diff |
|---|---|
| Registry accepts eager `{ Component }` + lazy `{ load }`; rejects mixed | ✅ `createRegistry.js:76-92` — both-set throws, non-function `load` throws, lazy-engine throws. Descriptor never mutated (test 12). |
| `mountApp` awaits `load()` on first mount, cached for subsequent | ✅ `mountApp.js:58-73` module-level `lazyAppCache` (per-id `React.lazy` wrapper) backed by `registry.resolveComponent` (per-id Promise cache, `createRegistry.js:138-180`). |
| System apps stay eager | ✅ `builtins.js:32-37` — `core:navigation` / `core:site-hub` / `core:toolbar-actions` / `core:notices-banner` / `core:notices-snackbar`. Matches plan list exactly. |
| Lazy chunks emit with named hints | ✅ `npm run build` → 25 `app-*.js` chunks (`app-appearance.js` 3.1K … `app-settings.js` 17K). `webpackChunkName` magic comments on every dynamic import. |
| Error handling for failed chunk loads | ✅ `mountApp.js:149-197` `AppErrorBoundary` only handles `ChunkLoadError` / `Loading chunk …` / `createRegistry: load() …` (via `isChunkLoadError`); render-time errors inside resolved components re-throw honestly. Retry clears both `lazyAppCache` and `registry.invalidateComponent` (webpack 5 does not auto-retry rejected `import()` promises). |
| `tests/runtime/registry-lazy-app.test.mjs` exists | ✅ 28 assertions covering both shapes, both-set rejection, cache identity, error paths, engine-lazy rejection, descriptor non-mutation, invalidate semantics. |
| Bundle size noted | ✅ `docs/v1-perf-baseline.md` Bundle-size table — `index.js` 2.06 MiB → 209 KiB (−90.1%); entrypoint 2.12 MiB → 250 KiB (−88%). |

## Tests (worktree)

```
lint:js        clean
test:schema    79 / 79
test:parity     4 / 4
test:runtime  359 / 359  (19 files; new lazy-app file contributes 28)
test:engines   77 / 77
build         clean — 25 app-*.js chunks emit; index.js 209 KiB; vendors chunk 245.js 1.75 MiB
```

PHP suites untouched by the diff — skipped re-run.

## Code review notes

1. **No `webpack.config.js` change needed** — plan flagged it as "possibly needed (publicPath, chunk-loading runtime)"; author confirmed via build that webpack 5's `publicPath: 'auto'` + `@wordpress/scripts` defaults handle chunk loading from `document.currentScript.src` of the boot `index.js`. CSS for lazy apps emits as sibling `build/app-<id>.css` and is auto-injected by MiniCssExtractPlugin's runtime — no PHP enqueue per-app. Nice — narrower scope than plan allowed.

2. **`Region.js` not touched** — plan said "Suspense boundary or equivalent." Author put the Suspense + boundary inside `mountApp.js` per-app instead of one Region-level boundary. Smaller blast radius (a load failure in one app doesn't crash the region) and aligns with the comment at `mountApp.js:265-284`. Correct call.

3. **`AppErrorBoundary` re-throws non-`isChunkLoadError` exceptions** (`mountApp.js:191-194`) — important: render-time crashes inside resolved apps surface honestly upstream rather than getting masked by "Failed to load app". This is explicitly noted both in the JSDoc and in the boundary code. Good.

4. **Kernel DS-neutrality preserved** — `AppLoadError`'s retry uses a plain `<button type="button">` (no `@wordpress/ui` Button) and `src/index.css` adds only structural rules (`min-block-size: 48px`, `padding: 16px`) with no DS tokens. Aligns with CLAUDE.md "Kernel is DS-neutral" rule.

5. **i18n correctness** — `AppLoadError` uses `__()` + `sprintf()` from `@wordpress/i18n` with a translator comment. Headline + retry button copy both wrapped.

6. **Module-level `lazyAppCache` is per-page, not per-registry** (`mountApp.js:58`). Today the kernel mounts one registry per page (shell switching is option-write + reload per `shell-switching.js`), so this is fine. If a future change ever supports in-page registry swap, this Map would carry stale `React.lazy` wrappers across registries. Not a real bug today — flagging for future-proofing if/when in-page shell switch arrives.

7. **`core:command-palette` is lazy** (`builtins.js:106-110`). Palette is `core:triggerable` (mounted hidden, opened via Cmd+K binding from `shells/*.json`). First Cmd+K invocation will trigger a `~1KB` chunk fetch before the palette renders — small but perceptible flicker. Worth deciding whether to promote to eager (zero-flicker, +1KB boot) or to add a preload hint when bindings register. Not a blocker for this PR; flag for browser smoke.

8. **Test file has a numbering typo** (`tests/runtime/registry-lazy-app.test.mjs:350`) — final block is labeled `// 13.` but should be `// 17.` (tests 13–16 were inserted above). Cosmetic; counts are correct (28 assertions pass).

9. **Test-count discrepancy** — PR body says "test:runtime — 350/350 (was 330; new file adds 20)" but actual runtime total is **359** (delta +29), of which the new file contributes **28**. CLAUDE.md correctly states 359 / 19 files. The PR body should be updated, but CLAUDE.md is the authoritative claim. Minor.

10. **Conflict surface — PR #42 (Track C / C4 dashboard widgets)** also touches `src/runtime/registry/builtins.js`. PR #42 adds three new entries (`core:dashboard-host`, `core:dashboard-widget-recent-posts`, `core:dashboard-widget-quick-draft`) to the *eager* `APP_COMPONENTS` map this PR renamed to `APP_LOADERS`. **Whichever merges second must translate the dashboard apps to the new lazy shape** (`{ load: () => import(/* webpackChunkName: "app-dashboard-host" */ '...') }`) — plan explicitly anticipated this in the coordination note. Heads-up for the merging reviewer.

11. **No scope drift.** 8 files touched, all within (and narrower than) plan's "Files touched" list. No recurring-pattern violations (no raw `fetch`, no eager DataViews import path change, no app-level a11y landmark drift).

## Gates remaining (you flagged these in the PR test plan)

- [ ] **Manual browser smoke — Network panel verification:**
  ```
  Open shell. Confirm boot loads `index.js` (~209 KiB) + `index.css` only.
  Navigate to `/posts` (or whichever DataViews-backed default route).
  Confirm a new `app-posts.js` request + a new `245.js` (vendors) request fires.
  Navigate to `/comments`. Confirm a new `app-comments.js` request fires
    but vendors chunk is cache-hit (no second download).
  ```
- [ ] **Force a chunk 404** (e.g. rename `build/app-posts.js` after boot) → verify inline `AppLoadError` renders, click Retry, restore the file, verify it recovers (validates `invalidateComponent` + `lazyAppCache.delete` path).
- [ ] **Cmd+K palette first-invocation smell test** — confirm first Cmd+K is acceptable UX after lazy migration; consider promotion to eager if flicker is jarring.

After the three smokes pass, squash-merge. **Suggested merge order: this PR first, then PR #42 (Track C / C4)** — #42 adds three new bundled apps that should register in lazy shape from the start; merging C4 first means D rebases over three new eager entries to flip them, which is more churn.

Issue #10, opened 2026-04-30
Tech-debt sweep: in-JS code dupes + stale comments


dabowman:
Small cleanups inside the M3 token system.

### 1. `compileSubtree()` duplicates main walk logic
`src/runtime/styles/compileStyles.js:66–86` — refactor to a shared traversal function with `compileStyles()`.

### 2. Duplicate WPDS-slot regex
`scripts/snapshot-wpds.mjs:36` and `tests/parity/wpds-snapshot.test.mjs:71` share an identical regex. Extract to a shared constant.

### 3. Stale comment on `parseColorToRgb`
`src/runtime/styles/compatBridge.js:46–70` — comment claims "small 3-line helper", function is 25 lines. Update or drop the comment.

Source: M3 review. Logged in `docs/feedback.md`.

dabowman:
Fixed in 635c78cf7916c5bc249e08b7947046d7e1b6ac1d. compileTree shared traversal, WPDS regex extracted to _slot-pattern.mjs, compatBridge docstring refreshed.

Issue #9, opened 2026-04-30
PHP/JS chrome-defaults seeded twice, byte-for-byte parallel


dabowman:
`src/runtime/config/normalizeV0.js:124–183` (`stylesFromBranding`) and `includes/origins/class-wp-admin-shell-origin-core.php:153–215` (`v0_styles_from_branding`) are mirrored seedings of chrome defaults from `branding.accentColor`.

Easy to drift — change one, forget the other, server-rendered config diverges from client. M4 added a parallel `__notices-banner` / `__notices-snackbar` region pinning that has the same drift surface (no new drift introduced in M4 but extends the affected area).

Options:
- Add a sync test that asserts the two outputs match for fixture inputs.
- Move the canonical default to a shared JSON file consumed by both sides.
- Pick one side as authoritative and have the other mirror the resolved styles tree from the JS bridge / PHP cascade output.

Source: M3 + M4 reviews. Logged in `docs/feedback.md`.

dabowman:
Resolved as a side effect of #4 (commit ee69abe). `src/runtime/config/normalizeV0.js` and its `stylesFromBranding` / `buildSystemApps` / `buildRegions` helpers were deleted; PHP `WP_Admin_Shell_Origin_Core::v0_styles_from_branding` and its sibling helpers are now the sole canonical defaults. No more drift surface.

Issue #8, opened 2026-04-30
Token-system lifecycle cleanup on shell-switch / unmount


dabowman:
Two lifecycle gaps in M3 token emission.

### 1. Density attr persists when new shell omits density
`src/runtime/styles/density.js` — if shell A declares `compact` and shell B omits `styles.density`, `data-wpds-density="compact"` may persist on `#wp-admin-shell`. Verify the default branch overwrites to `default` or removes the attr.

### 2. Token style tag never removed on kernel unmount
`src/runtime/styles/emitTokens.js` reuses `<style id="wp-admin-shell-tokens">` via `textContent` overwrite. No `.remove()` on unmount. Idempotent in normal use; multi-shell hot-reload could leave stale tokens between mounts.

Source: M3 review. Logged in `docs/feedback.md`.

dabowman:
Fixed in 8e8383b0ed3d2e29fc4eb2045bd985f9e460495d. v1 lifecycle verified clean (hard reload disposes DOM). clearTokens/clearDensity exposed for v2 in-process re-mount (issue #28).

Issue #7, opened 2026-04-30
Token compiler robustness: 8-digit hex, alias cycles, slot-path collision


dabowman:
Three corner-cases in the M3 token compiler.

### 1. Compat-bridge color parser rejects 8-digit hex + `rgba()`
`src/runtime/styles/compatBridge.js:46–70` — `parseColorToRgb()` handles `#fff`, `#3858e9`, `rgb()` only. If WPDS or theme.json resolves brand strong to `#3858e9ff` or `rgba(...)`, parser returns `null` and both `--wp-admin-theme-color--rgb` + `--wp-admin-theme-color-darker-20` silently drop. Compat bridge degrades without warning.

### 2. DTCG alias resolver has no cycle detection
`src/runtime/styles/compileStyles.js:122–147` — `resolveValue()` recurses through `{styles.path}` references with no visited-set or depth limit. Author config `a: '{styles.b}'` + `b: '{styles.a}'` causes stack overflow at runtime. Add visited-set + console warn + raw-string fallback.

### 3. Slot-path collision in `pathToWpds()`
`src/runtime/styles/compileStyles.js:114–115` — joins all path segments with `-`, so `a.bc` and `a-bc` both emit `--wpds-a-bc`. Author trust mitigates but no validation. Detect on emit + warn.

Source: M3 review. Logged in `docs/feedback.md`.

dabowman:
Fixed in b7ffd5d7800f166cec530236652ee80098338c9e. 8-digit hex + rgba parsing, alias cycle detection, slot collision warn. Production silent.

PR #41, opened 2026-05-14
feat(c1): REST preload as cascade primitive

dabowman:
## Summary

- Ships CIAB's REST preload middleware as a declarative `preload[]` block in admin.json (string paths or `[ path, method ]` tuples; GET / OPTIONS).
- New `WP_Admin_Shell_Preload` (under `includes/cascade/`) collects every cascade origin's `preload[]` through the existing `wp_admin_shell_data_{origin}` filters, concatenates additively, dedupes by `path|method`, hydrates via `rest_preload_api_request`, and emits `wp.apiFetch.createPreloadingMiddleware( ... )` as inline script on `wp-api-fetch` before the shell bundle enqueues.
- Cascade is additive only (no override semantics — preload entries carry no user-meaningful identity, so site/role/user simply append). Per-entry `Throwable` catch isolates failures so one bad path doesn't poison the bundle. Conditional preloads belong in a `wp_admin_shell_data_{origin}` filter callback.
- Spec §13 grows from 8 → 9 extension points; CLAUDE.md status block, file tree, extension-points list, and test counts updated.

Eliminates cold-mount round-trips `@wordpress/core-data` resolvers make for `/wp/v2/users/me`, `/wp/v2/types?context=view`, and the `__unstableBase` site fetch — biggest perf-per-LOC win in the CIAB-adoption plan.

Plan: `docs/plans/track-a-rest-preload.md`. Reference: `/wordpress/plugins/ciab-admin/lib/pages/admin.php:76` (CIAB's hardcoded list, the model for the schema'd version).

## Coordination notes

- `wp-admin-shell.php` `require_once` block: append-only — rebase if Track B / C land first.
- `docs/schemas/admin-v2.json`: adds top-level `preload` only — no conflict expected with Track C (which adds `dashboardWidget` to `admin-app-v2.json`).
- `CLAUDE.md` test-count line + extension-points numbering (#9): rebase if other tracks bumped these first. If B (#10/#11) or C (#12) land before this, renumber #9 on rebase.

PHP suite was not run from this worktree — wp-env binds 0.0.0.0:8888, which is owned by the parallel-tracks main checkout. Per `docs/plans/agent-prompts.md` "wp-env coordination", PHP suites are queued for the user (or designated agent on the main checkout) to run.

## Test plan

- [ ] Schema accepts string + tuple entries; rejects everything else (Ajv sweep 79 → 81 — positive `04-preload.json` + negative `07-preload-bad-shape.json`)
- [ ] `npm run test:schema` green
- [ ] `npm run test:runtime` green (no regressions)
- [ ] `npm run test:parity` green
- [ ] `npm run lint:js` clean
- [ ] PHP class collects, dedupes, and calls `rest_preload_api_request` (`run-preload-tests.php`, 20 assertions — needs `npx wp-env run cli wp eval-file ...`)
- [ ] Inline script lands before `wp-admin-shell` JS bundle (manual: view-source the shell page)
- [ ] Network panel: declared `/wp/v2/users/me` returns 200 from preloading middleware on first paint (no second request)
- [ ] CLAUDE.md test count updated to 865
- [ ] Spec §13 grows to 9 extension points


dabowman:
# PR Review — `feat(c1): REST preload as cascade primitive`

## Overview

New top-level `preload[]` block in admin.json. `WP_Admin_Shell_Preload` walks every cascade origin's `preload[]` via `wp_admin_shell_data_{origin}` filters, concat + dedupes on `path|method`, hydrates with `rest_preload_api_request`, emits inline `createPreloadingMiddleware` script on `wp-api-fetch` before shell bundle. CIAB-shape compat. 9th extension point. Schema fixture sweep +2, new PHP test file +20 → 865 total.

## Correctness

- **Filter double-fire risk.** `WP_Admin_Shell_Preload::collect_from_origins` re-runs `wp_admin_shell_data_{origin}` against raw `load_origins()` docs (resolver lines 105/116 also run them). Pure-functional filters: wasteful, no harm. Side-effecting filters (logging, registry mutation): fire twice per page render. Worth a comment in the class docblock explicitly stating filters must be idempotent, or — better — short-circuit by reusing already-filtered origins from `WP_Admin_Shell_Config::get()` cache.
- **Duplicate work.** `inject()` calls `load_origins()` independently of `WP_Admin_Shell_Config::get()`. Two full disk reads + origin loads per render. Cache layer (`WP_Admin_Shell_Cache`) probably hides cost, but worth verifying. Could plumb origins through from the surrounding `wp_admin_shell_render_page()` call.
- **`hydrate` GET form passes string path.** WP-core `rest_preload_api_request( $memo, $path )` does accept string for GET — confirmed. OK.
- **OPTIONS bucket shape.** `$cache['OPTIONS'][path]` matches WP-core gutenberg/edit-site preload convention. `createPreloadingMiddleware` parses correctly. Test `run-preload-tests.php:618` covers it.
- **Dedupe key `path|method`.** Pipe inside REST path effectively impossible. Fine.
- **`wp_script_is( ..., 'registered'|'enqueued' )` guard** (line 263) — correct; covers both states before enqueue order matters.
- **Inline-script timing.** Adds to `wp-api-fetch` at `'after'`. Works only if shell bundle declares `wp-api-fetch` as a dep so wp-api-fetch enqueues before the bundle. Not verified in diff — confirm in `wp_admin_shell_render_page()` / enqueue path. Manual smoke item #6 in PR description catches it via view-source, but a unit-level assertion would be better.

## Style / conventions

- PHP class location (`includes/cascade/`) matches sibling C2 view-config + field-collections classes. Consistent.
- `defined( 'ABSPATH' ) || exit;` present. `wp_json_encode` with `JSON_HEX_TAG | JSON_UNESCAPED_SLASHES` for inline script — appropriate. Consider adding `JSON_HEX_AMP` defensively (not strictly needed but matches WP-core preload code in `gutenberg_preload_api_request`).
- `Throwable` catch + `WP_DEBUG` `trigger_error` w/ `__()` — appropriate isolation per `Throwable` boundary.
- `normalize_entry` rejects malformed shapes silently (no `WP_DEBUG` warn). Class docblock says "warned under `WP_DEBUG`" — code doesn't actually warn on malformed entries (line 215 `continue` is silent). Either add the warn or update the comment.
- Test runner `$admin_id` (line 414) holds array of IDs — rename `$admin_ids` for clarity.
- Spec § 13 prose: long single-paragraph entry. Matches the C2 #7 entry's verbosity. Consistent w/ existing style.

## Performance

- Net win clear — eliminates 3 cold-mount RTs (`/users/me`, `/types?context=view`, `__unstableBase`). Server cost is one inline `rest_preload_api_request` per path; cheap on warm OPcache.
- Per-origin filter re-run (see Correctness) doubles filter callback execution per render. Quantify — if filters are pure, fine.

## Tests

- **PHP suite (20 assertions, `run-preload-tests.php`)**: covers `normalize_entry` happy + reject paths, multi-origin concat + dedup, distinct verbs, malformed neighbors don't poison, per-origin filter contribution, `hydrate` dispatch counter + cache shape (GET + OPTIONS bucket), empty-input safety. Good coverage.
- **Missing**: end-to-end `inject()` smoke that asserts `wp_add_inline_script` actually attaches. Could mock `wp_script_is` + capture via `$wp_scripts->print_inline_script`. Not blocking — manual smoke covers it.
- **Schema fixtures (+2)**: positive 04 covers string + tuple (GET + OPTIONS). Negative 07 covers assoc-shape only. Add negative for `["path", "POST"]` (bad verb) + `["path"]` (1-tuple). Schema's `enum: ["GET","OPTIONS"]` + `minItems: 2` would catch — pin them.
- **Filter idempotence test**: test only asserts filter ran once. Add an explicit case proving filter is invoked exactly once per origin during a `collect_from_origins` (rules out future double-fire regression).

## Security

- All preload paths route through `rest_preload_api_request`, which already enforces REST permissions per-request — no bypass. Failed permission → empty cache entry, client refetches and gets the normal 401. Good.
- Inline script JSON escape: `JSON_HEX_TAG` covers `</script>` injection. `JSON_UNESCAPED_SLASHES` keeps paths readable. Acceptable.
- No new REST endpoint added — preload is one-way emit only. Surface area minimal.
- `__()` text domain `'wp-admin-shell'` consistent w/ plugin convention.

## Risks

- **Coordination notes** in PR body call out `wp-admin-shell.php` require block ordering, schema add, extension-point #9 renumber. Confirmed append-only — no rebase landmines unless B/C land first.
- **PHP suite unrun from worktree** (wp-env binds main checkout). Queue for main runner per directive. Block merge on it.
- **Filter double-fire** is the only architectural concern. Low-impact for v1 callers (cascade filters in this repo are pure), but worth documenting.

## Recommended changes (priority)

1. **Class docblock** — explicitly state per-origin filters MUST be idempotent (double-fire reality).
2. **`normalize_entry` warn-or-comment** — either add `WP_DEBUG` warn on rejection or strike that promise from the class docblock.
3. **Schema negative fixtures** — add bad-verb + wrong-arity cases.
4. **Inject-time smoke** — assert inline script attaches (one assertion, ~10 lines).
5. **Verify `wp-api-fetch` dep chain** — confirm shell bundle has it; document in CLAUDE.md if not obvious.
6. **Rename `\$admin_id` → `\$admin_ids`** — cosmetic.

LGTM after #1 + #2 + PHP suite run. Other items follow-up-able.

dabowman:
Review addressed in 76f2266:

1. ✅ **Filter idempotence** — class docblock now states the contract explicitly: per-origin callbacks MUST be idempotent because the collector re-runs them independently of the resolver pass.
2. ✅ **`normalize_entry` warn promise** — struck the unmet WP_DEBUG line. Schema validation catches the same cases at authoring time, so a runtime warning would be redundant noise.
3. ✅ **Schema negatives** — added `08-preload-bad-verb.json` (POST tuple) + `09-preload-bad-arity.json` (1-tuple). Pins the verb enum + `minItems: 2`.
4. ✅ **Inject smoke** — new end-to-end test registers stub `wp-api-fetch`, seeds via filter, asserts both `createPreloadingMiddleware(` shape and preloaded path land on the `wp-api-fetch[after]` bucket.
5. ✅ **`wp-api-fetch` dep verified** — confirmed via `npm run build`: `build/index.asset.php` lists `wp-api-fetch` (transitively required by `@wordpress/core-data` + direct `apiFetch` usage in 5 apps). Shell bundle enqueues `wp-api-fetch` first → `wp_add_inline_script(..., 'after')` lands.
6. ✅ **Rename** — `$admin_id` → `$admin_ids`.

Counts: PHP 20 → 22, schema 81 → 83, total 865 → 869.

Deferred: short-circuit by reusing already-filtered origins from `WP_Admin_Shell_Config::get()` cache + `JSON_HEX_AMP`. Both follow-ups; cascade cache layer hides the double-load cost in practice.

dabowman:
LGTM. Scope matches `docs/plans/track-a-rest-preload.md`; tests green in the worktree.

## Verified against plan

| Plan item | Diff |
|---|---|
| Schema: string + tuple entries, top-level `preload[]` + `$defs/preloadEntry` | ✅ `docs/schemas/admin-v2.json` adds top-level `preload` array (lines 128–134 in diff) + `preloadEntry` definition with `oneOf` (string `^/.+` OR 2-prefix tuple, `items:false`, `min/maxItems:2`) |
| Tuple verb enum | ⚠️ Narrowed from plan's `GET\|POST\|PUT\|DELETE\|PATCH` to `GET\|OPTIONS` only. Justified inline ("CIAB convention preloads only GET + OPTIONS — REST controller dispatches OPTIONS for schema discovery"). Plan-text divergence is defensible; `rest_preload_api_request` doesn't usefully preload non-idempotent verbs. Flagging for awareness; no change requested. |
| `WP_Admin_Shell_Preload` collects + dedupes + injects on `wp-api-fetch` | ✅ `class-wp-admin-shell-preload.php`: `normalize_entry` → `collect_from_origins` (walks `ORIGINS_ORDER`, applies `wp_admin_shell_data_{origin}`, dedupes by `$path\|$method`) → `hydrate` (per-entry `Throwable` catch, OPTIONS bucket via `[path,'OPTIONS']` tuple form) → `inject` (registered-or-enqueued guard, empty-list short-circuit, `wp_add_inline_script(..., 'after')` with `wp.apiFetch.use(wp.apiFetch.createPreloadingMiddleware(...))`, `JSON_HEX_TAG \| JSON_UNESCAPED_SLASHES`) |
| Inline script lands BEFORE shell bundle | ✅ Attached to `wp-api-fetch[after]`, which is a transitive dep of `wp-admin-shell`. DOM emission order: wp-api-fetch tag → inline script → wp-admin-shell tag. Call-site ordering at `wp-admin-shell.php:270` (after `wp_enqueue_script('wp-admin-shell', ...)` at :255) is irrelevant — `wp_add_inline_script` queues; emission order is dep-graph driven. |
| Per-origin contribution via existing `wp_admin_shell_data_{origin}` filters | ✅ No new filter added. `collect_from_origins` calls `apply_filters("wp_admin_shell_data_{$origin}", $doc)` line :71. |
| `run-preload-tests.php` cascade merge / dedup / malformed-skip / real REST hit | ✅ 22 assertions covering: 11× `normalize_entry` cases (string, lowercase, missing slash, empty, int, assoc-shape, 1-tuple, 3-tuple, unknown verb, non-string method), single-origin preserve, multi-origin additive concat with cross-origin dedup, distinct-verbs-on-same-path both kept, malformed-entries-skip-neighbors-survive, per-origin filter contribution lands, `hydrate` actually fires `rest_pre_dispatch` + populates `/wp/v2/users/me` + `OPTIONS['/wp/v2/settings']` buckets, empty-list short-circuit, `inject` end-to-end inline-attached on `wp-api-fetch[after]`. |
| Positive + negative schema fixtures | ✅ Plan asked for 1 positive + 1 negative; author shipped 1 positive + 3 negative (`07-preload-bad-shape.json` assoc shape, `08-preload-bad-verb.json` POST, `09-preload-bad-arity.json` 1-tuple). Bonus coverage; no scope drift. |
| CLAUDE.md extension points #9 + test counts | ✅ Status block adds C1 paragraph; extension points 8 → 9; test surface 843 → 869; `npm run test:schema 79 → 83`; PHP runner list adds `run-preload-tests.php # 22`; file-tree adds `class-wp-admin-shell-preload.php`. |
| Spec §13 #9 description | ✅ `docs/wp-admin-shell-design-spec.md:891` "Eight" → "Nine"; new bullet 9 added with full semantics + CIAB compat note. |

## Tests (worktree)

```
lint:js          clean
test:schema      83 / 83
test:parity       4 /  4
test:runtime    all 18 files green (resolveRegion … view-config-hydrate-inline)
test:engines     77 / 77 (7+17+39+14)
build            clean (3 perf warnings, expected)
```

PHP suite (`run-preload-tests.php`, 22 assertions) deferred — wp-env is bound to the main checkout on this machine. Will run sequentially after merge.

## Code review notes

1. **Double-pass of `wp_admin_shell_data_{origin}` filters.** `WP_Admin_Shell_Preload::collect_from_origins` calls `apply_filters("wp_admin_shell_data_{$origin}", $doc)` (line :71) independently of `WP_Admin_Shell_Resolver::resolve_with` which runs the same filter chain (:105, :116). Two passes per render. Author flagged this in the class doc-block ("Filter idempotence requirement") and explained why it's unavoidable: the resolver's merge for `preload[]` would *replace* origin-wise (default array merge semantics), but preload semantics demand *additive concatenation* across origins. Reading off the resolved merged tree wouldn't give the union. Correct call. The mitigation is the doc-block warning to plugin authors that side-effecting filter callbacks (logging, REST registration) must be idempotent — acceptable cost for the cascade primitive.

2. **`inject()` short-circuits on missing `rest_preload_api_request` or unregistered `wp-api-fetch`** (lines :122, :125). Defensive — covers CLI / test contexts where `@wordpress/scripts` registration didn't run. Matches the test runner's `wp_register_script` stub-attach pattern at line :647.

3. **`hydrate` failure isolation via `Throwable` catch** (line :172). Single bad path doesn't poison the bundle; failed entries simply round-trip from the client. WP_DEBUG-gated `trigger_error` for diagnostics. Solid pattern.

4. **OPTIONS bucket key shape** — `rest_preload_api_request($cache, [$path, 'OPTIONS'])` lands the response under `$cache['OPTIONS'][$path]`, matching what `createPreloadingMiddleware` reads (mirrors gutenberg + edit-site preload shape). Verified in test runner line :628.

5. **Schema regex** `^/.+` correctly accepts query-string paths (`/wp/v2/types?context=view` in positive fixture). Matches plan's "query-string `_fields` filters and other GET parameters are part of the path string" guidance.

6. **No scope drift.** 10 files changed; 8 from plan's "Files touched" + 2 bonus negative fixtures (08, 09). All four sections of CLAUDE.md updates land cleanly; no collision with C2's recent merge (which added #7 + #8 — C1 appends #9 as planned).

7. **Coordination notes ack'd.** PR body explicitly flags the rebase-touchpoints (`wp-admin-shell.php` require block, CLAUDE.md test count, spec §13 numbering) per the plan's "Coordination" section, and is the first track to land after C2, so no actual conflicts.

8. **Author's empty-cache short-circuit at line :135** — `if ( empty( $cache ) ) return;` — guards against the case where every preload entry failed in `hydrate`. Defensible; means no broken inline script emitted on total failure.

## Gates remaining (you flagged these in the PR test plan)

- [ ] **PHP suite:** `npx wp-env run cli wp eval-file wp-content/plugins/WordPress-Admin-Environment/tests/php/run-preload-tests.php` — expect 22/22 (run sequentially after merge; wp-env single-stack constraint).
- [ ] **Browser smoke:** view-source the shell page, confirm `wp.apiFetch.createPreloadingMiddleware(...)` inline script lands between the `wp-api-fetch` script tag and the `wp-admin-shell` script tag.
- [ ] **Network panel:** declared `/wp/v2/users/me` returns 200 from preloading middleware (no real HTTP request) on cold mount.

Squash-merge after the PHP suite + browser smoke confirm.

Issue #2, opened 2026-04-30
Drawer region accessibility + effect deps


dabowman:
Two issues in `src/runtime/regions/drawer-region/index.js` from M1 review.

### 1. Missing dialog semantics (a11y)
Aside has only `data-region-id`. M5 a11y checklist (b) calls for `role="dialog"` + `aria-modal="true"` + `aria-labelledby` on overlay regions. Drawer should follow the same rule. Add focus trap on open + focus restore on close.

### 2. Keydown listener re-binds on every render (perf)
`src/runtime/regions/drawer-region/index.js:37` — `cfg.dismissOn` (object) in deps array, identity changes re-run the effect. `useMemo` over `normalizeDismiss(cfg.dismissOn)` cleaner.

Source: M1 review of commits `99a8627` + `8321f73`. Logged in `docs/feedback.md`.

dabowman:
Fixed in 4e28b14ba0e7723b6e35133cab32f2587d61177f. role/aria-modal/aria-labelledby + useFocusOnMount/Return/ConstrainedTabbing + memoized dismiss deps. Built clean.

PR #42, opened 2026-05-14
feat(c4): dashboard widget grid — first-class widget primitive

dabowman:
## Summary

Track C of the CIAB-adoption parallel tracks (`docs/plans/track-c-dashboard-grid.md`). Adds a first-class widget grid primitive: widgets are apps with a `dashboardWidget` manifest block; admin.json `dashboardWidgets` overrides position/size/visibility per id. CIAB's `next_admin_register_dashboard_widget()` ports to `wp_admin_shell_register_dashboard_widget()` and writes into the plugin origin via `wp_admin_shell_data_plugin`.

- New `core:dashboard-grid` region template (auto-fill CSS grid + declares `core:dynamic-children` for future engines that want to drive widget mounts as runtime-mutable child regions).
- New `core:dashboard-host` app: reads manifest registry + admin.json overrides via `composeWidgets()` and renders tiles via `<MountedApp>` — existing 4-layer cap gating + theming flows through unchanged.
- Two bundled example widgets: `core:dashboard-widget-recent-posts` (drafts list, ported from `core:dashboard`'s recent-drafts card) + `core:dashboard-widget-quick-draft` (Substack-style mini form; on submit creates a draft, invalidates the recent-drafts query, navigates to the full editor).
- PHP shim with two flavors: **override-only** (placement/size/hidden for an already-registered app) and **standalone** (synthesizes a minimal app manifest from `script` + `role` + `capabilities` so a mu-plugin can drop a widget in one call). admin.json wins on id collision (entry-replacement, same pattern as `fieldCollections`).
- Schema: `dashboardWidget` block in `admin-app-v2.json`, `dashboardWidgets` top-level in `admin-v2.json`, positive + negative fixtures for both.
- `/dashboard-grid` demo route in `shells/developer-admin.json` mounting `core:dashboard-host`. Legacy `/dashboard` + `core:dashboard` still ship — promotion deferred to a follow-up after stability per the plan.
- CLAUDE.md status block, app table, test counts; spec §13 #12.

Estimated 884 → **890 assertions** total. Browser smoke per the plan's acceptance criteria pending; CI runs the PHP / Node suites.

## Coordination notes

- `src/runtime/registry/builtins.js` adds three new bundled apps in the eager-render shape. Track D rewrites this file to the lazy `{ load: () => import() }` form on `feat/c5-lazy-app-loading`; merge order doesn't matter — either side trivially folds the other's additions.
- `shells/developer-admin.json` appends one route + one nav item. Track F sub-tracks may also touch this file; JSON merge.
- `docs/schemas/admin-v2.json`: Track A adds `preload` top-level; this adds `dashboardWidgets` top-level. Both append-friendly.
- `wp-admin-shell.php`: appends one `require_once`.
- `CLAUDE.md` status/app-table/test-counts: rebase pass per the parallel-tracks plan.

## Test plan

- [x] `npm run test:schema` — 86 PASS (adds 4 C4 fixtures + 3 bundled widget manifests).
- [x] `npm run test:runtime` — 348 PASS, 19 files (adds `compose-widgets.test.mjs`, 17 assertions).
- [x] `npm run test:parity` — 4 PASS.
- [x] `npm run test:engines` — 77 PASS.
- [x] `npm run lint:js` clean.
- [x] `npm run build` clean (1 pre-existing bundle-size warning — Track D's concern).
- [ ] `run-dashboard-widgets-tests.php` via wp-env (17 PHP assertions). Worktree wp-env conflicted with the main repo's running container on this machine; PHP file syntax-validated locally with `php -l`. Will run in CI on the PR.
- [ ] Browser smoke: navigate to `/dashboard-grid` in developer-admin, see two bundled widgets render in declared positions; register a third widget via the shim from a mu-plugin and confirm it appears; declare an admin.json override to hide one and confirm it disappears.


docs/wp-admin-shell-design-spec.md
```diff
@@ -899,6 +899,8 @@ Eight extension points, in increasing power:
...
...)` so a single PHP call covers both the manifest entry and the widget block — useful when a mu-plugin ships a widget without a full `apps/{name}/app.json` on disk. **Cascade resolution.** admin.json `dashboardWidgets[id]` wins over programmatic contribution on collision (entry-replacement, not per-property deep merge — same pattern as `fieldCollections`). The bundled host `core:dashboard-host` reads `window.wpAdminShell.manifests.apps` + the resolved `dashboardWidgets` block via `composeWidgets()` and renders surviving widgets as tiles; sizes translate to `grid-column: span N` / `grid-row: span M`, explicit positions to `grid-row/column-start`. Cap gating, theming, and DS-mismatch warnings flow through the existing 4-layer mount path because widgets are apps — there is no separate widget gating layer. The `core:dashboard-grid` region template ships with the `core:dynamic-children` platform service so a future engine compositor can drive widget mounts as runtime-mutable child regions, but the bundled host renders widgets directly via `<MountedApp>` for v1 simplicity. Drag-to-reorder, per-user widget ordering, and a wp-core dashboard-widget bridge are out of scope for this primitive.
```
dabowman: `entry-replacement, not per-property deep merge` describes only the PHP→admin.json layer (cascade contribution at `class-wp-admin-shell-dashboard-widgets.php:218`). The **manifest→admin.json** layer at JS render time is per-property (see `composeWidgets.mjs:90-101` — `override?.minSize ?? block.minSize`, etc.), and the schema `$defs/dashboardWidget` description confirms it ("admin.json wins per-property").

Two distinct merge layers with opposite rules; spec + CLAUDE.md conflate them. Suggest splitting into two sentences:

> admin.json `dashboardWidgets[id]` wins entirely over the PHP-registered override for the same id (entry-replacement, same pattern as `fieldCollections`). The resolved override then merges *per-property* over the app manifest's `dashboardWidget` block — admin.json winning each field.

Same fix needed in CLAUDE.md status block.

src/apps/dashboard-widget-quick-draft/index.js
```diff
@@ -0,0 +1,107 @@
...
+						context: 'edit',
+						orderby: 'modified',
+						order: 'desc',
+					},
+				] );
+				if ( draft?.id ) {
+					navigate( `#/posts/${ draft.id }/edit` );
+				}
+				setTitle( '' );
+				setContent( '' );
```
dabowman: `navigate()` on line 56 will hash-change → router will swap the route, unmounting the host 's tile and this component. The `setTitle('')` / `setContent('')` calls that follow then dispatch state on an unmounted component → React `setState on unmounted` warning.

Either drop the resets (component unmounts anyway), or reset before navigate:

```js
if ( draft?.id ) {
    setTitle( '' );
    setContent( '' );
    navigate( `#/posts/${ draft.id }/edit` );
}
```

Same `setSaving(false)` in the `finally` block hits this on the success path too — guard with a `mounted` ref, or short-circuit the `finally` when `draft?.id` is truthy.

src/apps/dashboard-widget-quick-draft/index.js
```diff
@@ -0,0 +1,107 @@
...
+					'postType',
+					'post',
+					{
+						per_page: 5,
+						status: 'draft',
+						context: 'edit',
+						orderby: 'modified',
+						order: 'desc',
+					},
+				] );
```
dabowman, line 52: The invalidation query shape duplicates `dashboard-widget-recent-posts`'s query (`per_page: 5`, `status: 'draft'`, `context: 'edit'`, `orderby: 'modified'`, `order: 'desc'`) verbatim. If recent-posts adds one filter (or changes `per_page`), this invalidate silently stops matching and the sibling tile goes stale — symptom only visible at runtime, no test would catch it.

Two fixes worth considering:
- Extract the query to a shared module both widgets import (`src/apps/dashboard-widget-recent-posts/query.js`), or
- Drop the shape-specific call and invalidate broader: `invalidateResolution( 'getEntityRecords', [ 'postType', 'post' ] )` — over-invalidates but stays correct as the recent-posts query evolves.

App docs already flag this in `app.md#invalidate-recent-drafts` — codifying it removes the runtime drift risk.

src/runtime/engines/core-default/index.css
```diff
@@ -154,6 +154,51 @@
...
+	flex-direction: column;
+	min-block-size: 0;
+	background: var(
+		--wp-admin-shell--chrome--content--card-background,
+		var(--wpds-color-bg-surface-neutral)
+	);
+	border: 1px solid var(--wpds-color-stroke-surface-neutral);
+	border-radius: var(--wpds-border-radius-md);
+	overflow: hidden;
+}
```
dabowman: Tile chrome (background / border / radius / header padding / body padding) styles host-emitted DOM (`wp-admin-shell-dashboard-tile`, `__header`, `__body` from `dashboard-host/index.js:93-102`) but lives in the core-default engine. A different engine (single-pane, desktop, plugin-shipped Material) gets unstyled tiles unless it copies these rules.

`dashboard-host/app.json#design-system-leakage` already declares the host 'DS-light', and CLAUDE.md states **bundled apps are WPDS-married by design** (DS-pluggability contract). The natural home for these rules is `src/apps/dashboard-host/index.css` next to the markup that emits the classes — same pattern PostsApp / CommentsApp / etc. use for their `@wordpress/ui`-flavored chrome.

The `[data-region-id$="dashboard-grid"] .wp-admin-shell-region__app { padding: 0 }` rule above (line 167) **is** correctly engine-side — it modifies engine-owned region wrapper padding. Keep that one here; move the three tile rules to the host app.

includes/cascade/class-wp-admin-shell-dashboard-widgets.php
```diff
@@ -0,0 +1,223 @@
...
+				$manifest_widget = $widget_block;
+				unset( $manifest_widget['hidden'] );
+				if ( ! empty( $manifest_widget ) ) {
+					$manifest['dashboardWidget'] = $manifest_widget;
+				}
+			}
+			self::$synthetic_manifests[ $id ] = $manifest;
+			// Forward to the manifest registry so the kernel's
+			// existing app pipeline picks it up. Errors propagate.
+			$registry_result = WP_Admin_Shell_Manifest_Registry::instance()->register_app( $manifest );
```
dabowman: `WP_Admin_Shell_Manifest_Registry::instance()->register_app( $manifest )` runs synchronously inside `register()`. What guarantees the registry is loaded by the time a mu-plugin / regular plugin calls `wp_admin_shell_register_dashboard_widget()`?

If a plugin author hooks the registration on `plugins_loaded` priority 0 (or in mu-plugin top-level), the shell's own `wp-admin-shell.php` may not yet have required `class-wp-admin-shell-manifest-registry.php`, and the call fatals on `Class "WP_Admin_Shell_Manifest_Registry" not found`.

C2's `wp_admin_shell_register_field_collection()` defers contribution to the `wp_admin_shell_data_plugin` filter (read-time, not register-time), which sidesteps the ordering problem entirely. Worth considering the same shape here:

- Stash the synthetic manifest in `self::$synthetic_manifests` (already happening).
- Forward to the manifest registry on `init`/`plugins_loaded` priority 20 instead of inline in `register()`.

Or, document the loading-order contract in the function's docblock so callers know what hook is safe.

includes/cascade/class-wp-admin-shell-dashboard-widgets.php
```diff
@@ -0,0 +1,223 @@
...
+			self::$overrides[ $id ] = $override;
+		}
+
+		// Standalone flavor: $args carries 'script' → synthesize a
+		// manifest. The script handle is required for an app the kernel
+		// can mount; without it, treat the call as override-only.
+		if ( isset( $args['script'] ) && is_string( $args['script'] ) ) {
+			$widget_block = isset( $args['dashboardWidget'] ) && is_array( $args['dashboardWidget'] )
+				? $args['dashboardWidget']
+				: $override;
```
dabowman: When a caller passes both `dashboardWidget => [...]` AND top-level placement keys (`defaultSize`, `position`, etc.) — perfectly valid against the public docblock — the function silently splits them:

- top-level keys → `$override` → contributed to admin.json cascade
- `$args['dashboardWidget']` → `$manifest['dashboardWidget']` (line 135)

So the synthetic manifest carries one set of sizes and the cascade carries another. `composeWidgets()` then per-property-merges them at render time — admin.json wins per-property — yielding a Frankenstein result the caller never intended.

Either:
- Document precedence + warn when both forms supplied, or
- Make `$widget_block` the merged form (`array_merge( $override, $args['dashboardWidget'] ?? [] )`) so one source of truth flows to both places.

Tests cover one or the other path, not the mixed call.

src/runtime/engines/core-default/engine.json
```diff
@@ -120,6 +121,20 @@
...
+			"default-style": {
+				"display": "grid",
+				"grid-template-columns": "repeat(auto-fill, minmax(280px, 1fr))",
+				"grid-auto-rows": "minmax(160px, auto)",
+				"gap": "var(--wpds-dimension-padding-lg)",
+				"inline-size": "100%",
+				"padding": "var(--wpds-dimension-padding-md)",
+				"align-content": "start"
+			}
+		},
```
dabowman, line 133: Template ships an auto-fill grid (`grid-template-columns: repeat(auto-fill, minmax(280px, 1fr))`, `gap`, `padding`, `align-content: start`) but the bundled host renders its **own** grid container with identical rules in `dashboard-host/index.css:1-8`. No bundled shell mounts a region with this template — `shells/developer-admin.json` puts `core:dashboard-host` inside the `content` region using `core:main`, so the template's grid CSS is unreachable today.

Three possibilities:
1. **Drop the template** until a compositor app actually populates regions via `useDynamicChildren`. Re-add it then.
2. **Use the template, drop the inline grid in the host** so mounting in this template is the supported path and the auto-fill rules live in one place.
3. **Keep the duplication** as forward-looking infra (current choice) — but at minimum, add a smoke test that mounts a region using the template and asserts the grid renders. Right now nothing exercises this code path; it ships untested.

Option 2 reads cleanest given the host is the only widget host in v1. Option 1 is honest about scope. Option 3 needs a test gate.

dabowman (commented):
Reviewed code paths: PHP shim + cascade contribution, composeWidgets + tests, host + widget components, schemas, engine template, demo route, spec/CLAUDE.md wording. Net: solid primitive, ships at the right size for v1. Six inline comments — one doc bug, one likely-fatal ordering risk, one runtime warning bug, three design questions.

**Must-fix before merge**
- **Doc inconsistency**: spec §13 #12 + CLAUDE.md describe a single "entry-replacement, not per-property" merge rule, but the code has two layers — PHP→admin.json (entry-replacement) and manifest→admin.json (per-property in `composeWidgets`). Schema's `$defs/dashboardWidget` already states the second rule correctly. (Inline comment on the spec.)
- **Manifest registry ordering**: `wp_admin_shell_register_dashboard_widget()` synchronously calls `WP_Admin_Shell_Manifest_Registry::instance()->register_app()` inside `register()`. A mu-plugin or early-`plugins_loaded` call before the shell's main file requires the registry class → fatal. C2 `wp_admin_shell_register_field_collection()` uses a filter-deferred contribution model that sidesteps this. (Inline on the PHP class.)
- **Quick-draft setState after navigate**: hash-change unmounts the component before the post-navigate state resets fire → "setState on unmounted" warning, and the `finally`'s `setSaving(false)` hits the same path. (Inline on the widget.)

**Worth resolving but not blocking**
- **Hardcoded invalidation query** in quick-draft duplicates recent-posts's exact query shape; one field drift on the consumer side silently breaks the cross-widget refresh. Suggest a shared constant or a broader `[ 'postType', 'post' ]` invalidate.
- **Standalone-flavor dual-source**: when callers pass both `dashboardWidget` and top-level placement keys, the values split between manifest and cascade — `composeWidgets` Frankenstein-merges them. Tests don't cover the mixed call.
- **Template/host grid duplication**: no shell mounts `core:dashboard-grid`. Either drop the template, drop the host's inline grid CSS, or test-gate the dead path.

**One DS-pluggability nit**
- Tile chrome (`.wp-admin-shell-dashboard-tile{,__header,__body}`) is defined in `core-default/index.css` but styles host-emitted DOM. Alt engines get unstyled tiles. Per CLAUDE.md's "bundled apps are WPDS-married", the natural home is `src/apps/dashboard-host/index.css`. The `__app { padding: 0 }` scoped rule rightly stays engine-side.

**Solid bits worth calling out**
- The host's "manifest is the eligibility check" design is clean — widgets are apps, cap gating and theming flow through the existing 4-layer mount path with no new gating layer. Schema, PHP, JS, and tests all align on this.
- `composeWidgets` is a properly pure ESM helper with thorough fixture coverage (17 assertions covering eligibility, defaults, hidden, per-property override, clamping, malformed inputs, ordering). Good shape.
- `data-region-id$="dashboard-grid"` engine scoping for the `padding: 0` rule is the right pattern — shell-id-agnostic, works for multi-grid shells.

dabowman:
Addressed all six review comments + the DS-pluggability nit in `7e8e4b0`.

**Must-fix**:
- ✅ **Doc inconsistency** — spec §13 #12 + CLAUDE.md now describe both layers explicitly: admin.json wins entry-replacement over the PHP override, then resolves per-property over the manifest's `dashboardWidget` block via `composeWidgets`. Schema $defs already correct.
- ✅ **Manifest-registry ordering** — synthetic manifests now stash into `$pending_registrations` and flush at `init` priority 7 (lazy-flushed earlier on first `wp_admin_shell_data_plugin` apply via `flush_pending_registrations()`). Mirrors the field-collections deferral pattern. Mu-plugin / early-`plugins_loaded` callers fire safely. New test asserts the pre-flush registry state is empty + the lazy filter-flush works.
- ✅ **Quick-draft setState after navigate** — success path now navigates and returns immediately; `setSaving(false)` only fires on the error early-return and the defensive no-draft-id branch. No `finally` block.

**Worth resolving**:
- ✅ **Hardcoded invalidation query** — extracted `RECENT_DRAFTS_QUERY` to `src/apps/dashboard-widget-recent-posts/query.mjs`. Both widgets import the same frozen object so future query changes can't silently drift.
- ✅ **Standalone-flavor dual-source** — `register()` now merges top-level placement keys over `args['dashboardWidget']` (top-level wins per-property) into one resolved override that flows to both the cascade contribution + the synthetic manifest. New test asserts the merge.
- ✅ **Template/host grid duplication** — went with option 2 (template carries declaration, host owns layout). Dropped the auto-fill grid CSS from the `core:dashboard-grid` template — now just `display: block; inline-size: 100%; block-size: 100%; overflow: auto` plus the `core:dynamic-children` declaration. Host's inline grid CSS remains the single layout authority.

**DS-pluggability nit**:
- ✅ Moved `.wp-admin-shell-dashboard-tile{,__header,__body}` chrome from `core-default/index.css` to `src/apps/dashboard-host/index.css`. Engine-side rule narrowed to the `region__app { padding: 0 }` scoped on `data-region-id$="dashboard-grid"` (touches engine-owned region-wrapper padding).

Test surface: **890 → 898 assertions**. Schema, lint, build clean. PHP `run-dashboard-widgets-tests.php` now 17 → 25 assertions covering the deferral pattern + dual-source merge.

docs/wp-admin-shell-design-spec.md
```diff
@@ -899,6 +899,8 @@ Eight extension points, in increasing power:
...
...ade resolution.** Two distinct merge layers stack: (1) admin.json `dashboardWidgets[id]` wins entirely over the PHP-registered override for the same id (entry-replacement, same pattern as `fieldCollections`). (2) The resolved override then merges *per-property* over the app manifest's `dashboardWidget` block at render time — admin.json winning each field via `composeWidgets()`. The bundled host `core:dashboard-host` reads `window.wpAdminShell.manifests.apps` + the resolved `dashboardWidgets` block and renders surviving widgets as tiles; sizes translate to `grid-column: span N` / `grid-row: span M`, explicit positions to `grid-row/column-start`. Cap gating, theming, and DS-mismatch warnings flow through the existing 4-layer mount path because widgets are apps — there is no separate widget gating layer. The `core:dashboard-grid` region template ships with the `core:dynamic-children` platform service so a future engine compositor can drive widget mounts as runtime-mutable child regions, but the bundled host renders widgets directly via `<MountedApp>` for v1 simplicity. Drag-to-reorder, per-user widget ordering, and a wp-core dashboard-widget bridge are out of scope for this primitive.
```
dabowman: Numbering: list runs 1-8, new entry numbered `12.`. Spec heading 9 lines above still reads "Eight extension points, in increasing power" (line 891). Should renumber the entry to `9.` and bump the header to "Nine".

Same issue in CLAUDE.md status block — header bumped to "Nine extension surfaces" but the new bullet stayed at `12.`. Pick one numbering scheme + apply to both docs.

dabowman (commented):
Re-review of `7e8e4b0` — fix commit addresses every prior item correctly. One small numbering nit (left inline), nothing else worth raising.

**Verified fixed**
- **Doc layer split**: spec §13 #12 and CLAUDE.md now both spell out the two-layer rule explicitly — admin.json wins entry-replacement over the PHP-registered override; the resolved override then merges per-property over the manifest's `dashboardWidget` block via `composeWidgets`. Schema $defs description aligns. ✅
- **Registry ordering**: `register()` now stashes into `self::$pending_registrations`. Flush has two paths: `add_action('init', ..., 7)` (runs before the shell's `init` priority-8 manifest discovery) and a lazy in-filter flush inside `wp_admin_shell_data_plugin`. `flush_pending_registrations()` is idempotent (early-return when empty, drains then clears). `class_exists` guard handles the edge case. mu-plugin / early-`plugins_loaded` callers safe. ✅
- **Quick-draft setState**: navigate-then-return on the success path; error path early-returns after `setSaving(false)` (no `finally`); resets only on the defensive no-draft-id branch. No more setState-on-unmounted. ✅
- **Invalidate single source of truth**: `RECENT_DRAFTS_QUERY` extracted to `src/apps/dashboard-widget-recent-posts/query.mjs`, `Object.freeze`d, imported by both widgets. Bonus: recent-posts dropped its `useMemo(() => ({...}), [])` for a direct constant import — stable ref across renders without the hook. ✅
- **Dual-source merge**: `register()` now seeds `$override` from `$args['dashboardWidget']` then overlays top-level keys, so one resolved doc flows to both `self::$overrides[$id]` and `$manifest['dashboardWidget']`. Two new tests pin the top-level-wins-per-property behavior + assert the synthetic manifest carries the merged value. ✅
- **Tile CSS relocation**: `.wp-admin-shell-dashboard-tile{,__header,__body}` rules moved to `dashboard-host/index.css`; engine retains only the `region__app { padding: 0 }` rule scoped on `data-region-id$="dashboard-grid"` with a comment explaining why it belongs engine-side. DS-pluggability contract preserved. ✅
- **Template scope**: `core:dashboard-grid` template trimmed to `display: block; inline-size: 100%; block-size: 100%; overflow: auto` + the `core:dynamic-children` declaration. Host owns the grid; template is honest forward-looking infra. ✅
- **Tests +8**: deferred-flush via init + lazy filter-flush + dual-source top-level-wins-per-property + synthetic-manifest single-source-of-truth assertion. 17→25. ✅

**One small nit (inline)**
- Numbering in spec §13 + CLAUDE.md extension-points list jumps 8 → 12. Spec heading still reads "Eight extension points" (CLAUDE.md heading was bumped to "Nine" but the bullet kept `12.`). Pick `9.` everywhere or leave `12.` and bump the spec heading to match.

Net: LGTM after the numbering pass. No blockers.

dabowman:
Renumbered C4 extension-point bullet `12.` → `9.` across spec, CLAUDE.md, PHP class/test docblocks, `composeWidgets.mjs` header, and the `admin-v2.json` schema description. Spec heading already bumped to "Nine extension points"; CLAUDE.md heading already "Nine extension surfaces" — consistent now. Plan docs (`track-*.md`, `agent-prompts.md`) keep their reserved-slot bookkeeping for parallel-track coordination.

Bonus cleanup: `4e70bd2` accidentally staged the worktree's `node_modules` symlink. Untracked in `e026ab3` + tightened `.gitignore` to match both `node_modules` (file/symlink) and `node_modules/` (directory).

dabowman (commented):
Re-review of `4e70bd2` + `e026ab3` — both clean, LGTM.

**`4e70bd2` — renumber 12 → 9**
- Spec §13 heading: "Eight" → "Nine" ✅
- Spec entry: `12.` → `9.`, both internal `spec §13 #12` refs updated to `#9` ✅
- CLAUDE.md status block + ext-points list: `#12` → `#9`, entry renumbered ✅
- `docs/schemas/admin-v2.json#dashboardWidgets.description`: `#12` → `#9` ✅
- PHP class header + public API docblock + tests header + `composeWidgets.mjs` header: all four `spec §13 #12` strings → `#9` ✅
- Plan-side `track-*.md` / `agent-prompts.md` reserved-slot bookkeeping left intact per the commit message. Reasonable scope.

**`e026ab3` — untrack node_modules symlink**
- `.gitignore` adds the bare `node_modules` pattern alongside the existing `node_modules/` directory pattern so worktree symlinks at that path are also ignored ✅
- One-line scope; no other drift ✅

Nothing else changed. Branch is in good shape to merge.

dabowman:
LGTM. Net: solid primitive, scope-disciplined, ships at the right size for v1. Prior review thread already chased down doc/ordering/setState/dual-source/template-grid items; final state on `e026ab3` is clean. Independent fresh-eyes pass below.

## Verified against plan

| Plan item | Diff |
|---|---|
| `core:dashboard-grid` template in `core-default/engine.json` | ✅ `engine.json:124` — `role: "region"`, `platform.core:dynamic-children: true`, minimal `default-style` (template owns declaration, host owns layout) |
| Two bundled widgets (`recent-posts` + `quick-draft`) each in own `src/apps/<id>/` dir w/ `index.js` + `app.json` + `app.md` | ✅ Both present; `recent-posts` also ships `query.mjs` (shared frozen query) |
| `wp_admin_shell_register_dashboard_widget()` public function | ✅ `wp-admin-shell.php:84` require + `class-wp-admin-shell-dashboard-widgets.php:510` shim. Two flavors (override-only + standalone). Synthetic-manifest forwarding deferred to `init` p7 + lazy filter-flush (mu-plugin / early-`plugins_loaded` safe) |
| admin.json `dashboardWidgets` can hide / resize / reposition | ✅ Schema $defs `dashboardWidgetOverride` carries `title?` / `defaultSize?` / `minSize?` / `position?` / `hidden?`; `composeWidgets` honors all four |
| Cap gating flows through app contract (no new gating layer) | ✅ Tiles mount via `<MountedApp>` → existing 4-layer gate fires. Each widget app declares `capabilities: ["edit_posts"]`; host has `additionalProperties: false` config-schema |
| `dashboardWidget` block on `admin-app-v2.json` w/ pos + neg fixtures | ✅ `admin-app-v2.json:272` $def + fixtures `app/positive/06`, `app/negative/08` |
| `dashboardWidgets` top-level on `admin-v2.json` w/ pos + neg fixtures | ✅ `admin-v2.json:128` block + fixtures `admin/positive/04`, `admin/negative/07` |
| `run-dashboard-widgets-tests.php` w/ registry CRUD + cascade override | ✅ 25 assertions covering validation, override flavor, cascade contribution, admin.json wins, standalone flavor, deferred-flush via init, lazy filter-flush, dual-source merge, override-only does not synthesize |
| Spec §13 #9 added + CLAUDE.md status/app-table/test-counts | ✅ Spec heading bumped to "Nine", §13 entry #9; CLAUDE.md status block, three new app-table rows, test count `843 → 898` |
| `/dashboard-grid` demo route in a bundled shell | ✅ `shells/developer-admin.json:211` (mounts `core:dashboard-host`); nav item at `:28` |

## Tests (worktree)

```
lint:js          clean
test:schema      86 / 86       (+7 from main: 4 C4 fixtures + 3 bundled widget manifests)
test:runtime    348 / 348      (19 files; new compose-widgets test = 17 assertions)
test:parity       4 / 4
test:engines     77 / 77
build           clean (3 pre-existing webpack performance warnings unchanged)
```

PHP `run-dashboard-widgets-tests.php` not executed locally (worktree wp-env conflicts with main repo's container per author's PR note); 25 assertions inline-readable in the diff, syntax-validated.

## Scope discipline (cross-cutting)

- 31 files touched, all within plan's "Files touched" list. No drift.
- **`core:dashboard` preserved** alongside new `core:dashboard-host` — exactly per plan option (b). `/dashboard` + `/dashboard-grid` coexist on developer-admin.
- **No drag-to-reorder UI** — out per plan; widget order is config-driven via `position`.
- **No `wp_dashboard_setup` bridge** — out per plan.
- **Uses existing `core:dynamic-children`** declaratively (template platform service) but host renders widgets directly via `<MountedApp>` — explicit in `dashboard-host/index.js:14-19` header comment ("v1 host is config-driven; future engine can replace with a compositor"). Honest forward-looking infra.

## Conflict surface (parallel tracks)

- `src/runtime/registry/builtins.js`: appends 3 imports + 3 entries to the eager `APP_COMPONENTS` map. **Track D (PR #40)** rewrites this to lazy `{ load: () => import() }`. Trivial fold either way — both shapes are append-friendly.
- `docs/schemas/admin-v2.json`: appends `dashboardWidgets` top-level. **Track A** appends `preload` top-level. Both clean appends, no key collision.
- `docs/schemas/admin-app-v2.json`: adds `dashboardWidget` $def. No collision w/ A/B/D.
- `shells/developer-admin.json`: appends 1 nav item + 1 route. **Track F sub-tracks** may also touch. JSON object merge, no collisions.
- `wp-admin-shell.php`: appends 1 `require_once`. Append-only.
- `CLAUDE.md`: status block, app-table rows, test-count line. Rebase pass per parallel-tracks plan.

## Code review notes

1. **`dashboard-host/index.js:43-47` useMemo dep array** — only watches `config?.dashboardWidgets`, not `window.wpAdminShell?.manifests?.apps`. Defensible because the manifests inline-script is page-lifetime immutable (`wp-admin-shell.php:328` snapshot at render time). Inline comment captures the reasoning. ✅
2. **Dead-code engine rule** — `core-default/index.css:163` `[data-region-id$="dashboard-grid"]` selector matches region IDs ending in "dashboard-grid", but `developer-admin.json` mounts `core:dashboard-host` into the existing `content` region (region ID `content`, not `dashboard-grid`). The rule only fires when a shell actually uses the `core:dashboard-grid` template + names its region accordingly — which no bundled shell does today. Not a bug (it's correctly scoped for when a shell does), but the rule is exercised by zero code paths until then. Worth a follow-up to either land a shell using the template or trim the rule.
3. **`composeWidgets.mjs:62-72`** — `pickPosition` falls back to `auto` for any non-`auto` non-`{row,col}` value (including malformed objects). Schema rejects `row: 0` at validation time; `composeWidgets` defends at runtime. Belt + suspenders, correct.
4. **`composeWidgets.mjs:84` clampToMin** — `defaultSize` floored to `minSize` per-axis. Behavior matches `app.md:46` rebuild-guide spec.
5. **`quick-draft/index.js:38-72`** — async submit handler post-fix: success path navigates + returns (no setState after unmount); error path early-returns after `setSaving(false)`; defensive no-id branch resets in-place. No `finally`. Comment on lines 60-64 explains the unmount race. ✅
6. **`recent-posts/index.js:16` + `quick-draft/index.js:14`** — both import the same frozen `RECENT_DRAFTS_QUERY` from `query.mjs`. Single source of truth; future query-shape changes can't silently drift the cross-widget invalidation. ✅
7. **Manifest registration deferral** — `register()` stashes into `$pending_registrations` instead of calling `WP_Admin_Shell_Manifest_Registry::instance()->register_app()` synchronously. Two flush paths: `add_action('init', flush_pending_registrations, 7)` (before main shell `init` priority 8) + lazy in-filter flush inside `wp_admin_shell_data_plugin`. `class_exists` guard. Mirrors C2 field-collections pattern. mu-plugin / early-`plugins_loaded` callers safe. ✅
8. **`composeWidgets.mjs:103` title fallback chain** — `override.title || block.title || manifest.title || id`. Schema enforces `minLength: 1` for `title` so empty-string can't sneak past; `||` (not `??`) is correct here because empty-string should fall through.
9. **`dashboard-host/app.json:36` `documentation.constraints.no-runtime-mutation`** — explicit "host is config-driven; widgets are computed once per (manifests, overrides) tuple." Drag-to-reorder explicitly disclaimed. Matches plan. ✅

## Gates remaining (per PR test plan)

- [ ] Browser smoke: navigate `/dashboard-grid` in developer-admin → two bundled widgets render in declared positions
- [ ] Register a third widget via the shim from a mu-plugin → confirm appears
- [ ] Admin.json override hides one of the bundled widgets → confirm disappears
- [ ] `run-dashboard-widgets-tests.php` via wp-env in CI (worktree-blocked locally)

Squash-merge once browser smokes + CI PHP suite are green. Track C is largest of the parallel tracks and the implementation is clean — no follow-ups needed beyond the dead-code engine rule (#2 above) which is harmless for the v1 cut.

Issue #11, opened 2026-04-30
Settings panel merge can shadow filtered-out builtin


dabowman:
`src/apps/SettingsApp.js:87–93` filters `BUILTIN_PANELS` by `config.panels[]` allowlist then concats `slotPanels` unconditionally.

Failure mode: if a shell narrows to `panels: ['general']` and a plugin registers a `core:settings.panels` item with `id: 'writing'`, the plugin panel renders even though the shell intended to hide it. A plugin can also register `id: 'general'` and produce two "General" entries in the sidebar.

Fix: dedupe by id + apply the allowlist to slot panels too. Document plugin-panel ID uniqueness in the slot contract either way.

Source: M4 review. Logged in `docs/feedback.md`.

dabowman:
Fixed in 07fff28163e43b154737c572ca5fbd4101376220. Allowlist applied to plugin panels, builtin-id collisions rejected, plugin-vs-plugin duplicates last-write-wins.

PR #43, opened 2026-05-14
feat(c3): menu-item + admin-route shims (CIAB compatibility)

dabowman:
## Summary

Track B of the [CIAB-adoption parallel tracks](docs/plans/ciab-adoption-tracks.md). Two PHP-side shims so plugins fluent in CIAB's `next_admin_register_menu_item()` + `next_admin_register_admin_route()` port to the shell with mechanical `s/next_admin_/wp_admin_shell_/g` rename.

- `wp_admin_shell_register_menu_item( \$id, \$args )` — args mirror CIAB 1:1 (`to` / `label` / `icon` / `badge` / `parent` / `parent_type` / `position`); shell adds optional `region` (defaults to first `core:navigation` region in the resolved tree, accepts bare ids or slash-paths) + optional `capability` (CIAB plugins did inline `current_user_can()` checks; the shell drops those — `capability` flows through the 4-layer cap model).
- `wp_admin_shell_register_admin_route( \$path, \$args )` — collapses CIAB's positional `($path, $content_module, $route_module, $before_load, $static_data, $gc_time)` into `($path, [ 'app' => …, 'config' => […], 'static_data' => […], 'gc_time' => … ])`. `app` replaces `content_module`; `static_data` folds into `config`; `gc_time` is accepted-and-ignored (TanStack-specific, no shell equivalent — emits a per-path `WP_DEBUG` notice).
- Both contribute through `wp_admin_shell_data_plugin` at priority 5 (mirrors `WP_Admin_Shell_Field_Collections`) so admin.json + plugin-author filters at default priority 10 still win on top.
- `parent_type=dropdown` falls back to `drilldown` with a per-id `WP_DEBUG` notice (shell nav has no dropdown primitive today).
- Duplicate ids/paths return `WP_Error`. No `unregister` — admin.json + the cascade are the override mechanism.

## Files

**New**
- `includes/cascade/class-wp-admin-shell-menu-items.php` (registry + nav-region resolver + filter contribution)
- `includes/cascade/class-wp-admin-shell-admin-routes.php` (registry + filter contribution)
- `tests/php/run-menu-route-shims-tests.php` (57 assertions)

**Modified**
- `wp-admin-shell.php` — `require_once` block + free function definitions
- `CLAUDE.md` — extension points #9 + #10, test counts (PHP suite total: 368 → 425; also fixed stale view-config drift 44 → 54)
- `docs/wp-admin-shell-design-spec.md` — §13 grew from 8 to 10 extension points

## Test plan

- [x] Schema sweep (`npm run test:schema`): 79/79 — no schema additions, additive contribution into existing blocks
- [x] Runtime tests (`npm run test:runtime`): pass — no kernel surface touched
- [x] Parity (`npm run test:parity`): 4/4
- [x] Engines (`npm run test:engines`): 7/7
- [x] Lint (`npm run lint:js`): clean
- [x] PHP suite (run-menu-route-shims-tests.php): 57/57
  - Both APIs reject malformed input + duplicates with `WP_Error`
  - Default region resolution + explicit region (bare id + slash-path) routing
  - Drilldown nesting via `parent` + `parent_type=drilldown`
  - Dropdown→drilldown fallback (no fatal)
  - `gc_time` accepted, never propagated to resolved route doc
  - Position sorting (null sorts last)
  - External-link auto-detection (https / protocol-relative / hash)
  - Empty registry → contribute is a no-op
  - End-to-end via `apply_filters('wp_admin_shell_data_plugin', \$bare_admin_json)` — both items + routes appear in resolved tree
  - Cap-collector picks up shim-declared `capability` so the 4-layer cap model gates shim items the same as inline admin.json items

## Coordination

- `wp-admin-shell.php` `require_once` block: append-only — trivial merge with parallel tracks
- Spec §13 numbering: this PR adds #9 (menu) + #10 (admin-route). Track A (REST preload) plans to add a #9 of its own. **If Track A merges first, rebase will renumber this PR's entries to #10 + #11.** Whichever lands second resolves the merge by bumping numbers + updating the §13 prose ("Ten extension points" → "Eleven").
- CLAUDE.md test-count line: rebase against whatever Track A or others land first.


includes/cascade/class-wp-admin-shell-menu-items.php
```diff
@@ -0,0 +1,482 @@
...
+		if ( ! empty( $item['icon'] ) ) {
+			$shell['icon'] = (string) $item['icon'];
+		}
+		if ( array_key_exists( 'badge', $item ) && $item['badge'] !== null ) {
+			$shell['badge'] = $item['badge'];
+		}
+		if ( ! empty( $item['capability'] ) ) {
+			$shell['capability'] = (string) $item['capability'];
+		}
+		if ( $item['external'] === true || self::is_absolute_url( $item['to'] ?? '' ) ) {
```
dabowman: `external` opt-out impossible. `$item['external'] === true || is_absolute_url($item['to'])` means setting `external => false` on an absolute-URL `to` still flags external. CIAB had no opt-out either, so not a regression — but if intent is "user wins over auto-detect," check `=== false` first. Minor; flag if explicit-false should win.

includes/cascade/class-wp-admin-shell-menu-items.php
```diff
@@ -0,0 +1,482 @@
...
+			$shell['external'] = true;
+		}
+		return $shell;
+	}
+
+	private static function to_shell_screen( $item, $kids ) {
+		$shell = array(
+			'screen' => (string) $item['id'],
```
dabowman, line 359: Screen id reuses registration id verbatim. Navigation app reads/writes `?screen=<id>` URL slot, so id must be URL-safe. Plus collision risk: inline admin.json `screen: 'settings'` + shim-registered `screen: 'settings'` resolve to the same `?screen=settings` slot. Two options: (a) document the constraint in the registry docstring, or (b) namespace shim screens (`plugin:<id>`). Lean (a) for CIAB parity.

includes/cascade/class-wp-admin-shell-menu-items.php
```diff
@@ -0,0 +1,482 @@
...
+				$children_of[ $parent ][] = $item;
+			} else {
+				$roots[] = $item;
+			}
+		}
+
+		$build = function ( $item ) use ( &$build, &$children_of ) {
+			$kids_input = $children_of[ $item['id'] ] ?? array();
+			$kids       = array_map( $build, $kids_input );
+			$is_screen  = ! empty( $kids ) || ( $item['parent_type'] ?? null ) === 'drilldown';
```
dabowman, line 321: Doc/impl drift. Docstring (lines 19-20 + spec §13#9) says `parent_type=drilldown` triggers screen emission. Impl says `! empty($kids) || parent_type === 'drilldown'` — so a parent without explicit `parent_type` becomes a screen as soon as anything references it as `parent`. Probably fine and ergonomic, but either align spec ("screen when parent_type=drilldown OR another item references this id as parent") or tighten impl. Test on line 270-280 only exercises explicit `parent_type=drilldown`.

includes/cascade/class-wp-admin-shell-menu-items.php
```diff
@@ -0,0 +1,482 @@
...
+				}
+				$region_path = $default_region_id;
+			} else {
+				$region_path = self::find_region_path( $doc, $region_id );
+			}
+			if ( ! $region_path ) {
+				continue; // No matching region in this shell — silently drop.
+			}
+			$by_region[ $region_path ][] = $item;
+		}
```
dabowman, line 261: Cross-region parents become roots. `build_tree` runs per-region bucket: if item A targets `sidebar/nav` and item B (parent=A) targets `sub-shell/inner-nav`, B lands as a root in `inner-nav` because A isn't in B's bucket's id_set. Silent. Either: (a) drop cross-region children with a `WP_DEBUG` notice, or (b) require parent + child to share a region (validate at register time). Edge case; CIAB had no cross-region nav, so plausibly fine to ignore + document.

includes/cascade/class-wp-admin-shell-admin-routes.php
```diff
@@ -0,0 +1,184 @@
...
+		if ( isset( $args['static_data'] ) ) {
+			if ( ! is_array( $args['static_data'] ) ) {
+				return new WP_Error(
+					'wp_admin_shell_admin_route_invalid_static_data',
+					__( 'Admin route "static_data" must be an array when provided.', 'wp-admin-shell' )
+				);
+			}
+			// CIAB pass-through. Fold into `config` so the admin-v2 schema
+			// (which only allows `app` + `config` on a route) accepts it.
+			$config = array_merge( $config, $args['static_data'] );
```
dabowman: Merge direction worth confirming. `array_merge($config, $args['static_data'])` lets `static_data` overwrite `config` keys on collision. CIAB convention treated `static_data` as preloaded route state (base); inline `config` was usually the override-y stuff. May want the inverse: `array_merge($static_data, $config)` so explicit `config` wins. Probably never collides in practice — flag for awareness.

dabowman (commented):
Solid shim work. Test surface is thorough (57 assertions cover the contribute path nicely). A few inline items — mostly small semantics gaps and one doc/impl drift in `build_tree`. None block; flag them now so the API contract is set before plugin authors lean on it.

Overall feel: 
- `wp_admin_shell_data_plugin` priority 5 is the right hook — plugin doc carries the regions tree, so default-region resolution works in production (not just tests).
- E2E test demonstrates filter wiring works alongside field-collections (also priority 5). Good.
- Schema-validation: this PR adds nothing to `admin-v2.json`. Emitted shapes (screen + link items, route entries) match what admin.json already produces, so validation passes incidentally. Worth re-verifying if/when schema-strict validation runs against the post-filter doc.

wp-admin-shell.php
```diff
@@ -156,6 +158,46 @@ function wp_admin_shell_register_shell( $slug, $admin_json ) {
...
+ * shell's 4-layer cap model. CIAB args (`to`, `label`, `icon`, `badge`,
+ * `parent`, `parent_type`, `position`) carry over 1:1; the shell adds
+ * an optional `region` arg (defaults to the first `core:navigation`
+ * region in the resolved tree).
+ *
+ * @param string $id   Menu-item id (must be unique within the registry).
+ * @param array  $args Args. See `WP_Admin_Shell_Menu_Items::register`.
+ * @return string|WP_Error Id on success, WP_Error on failure.
+ */
+function wp_admin_shell_register_menu_item( $id, $args ) {
```
dabowman, line 191: No timing note in docstring. Registry collects whenever called, but the `wp_admin_shell_data_plugin` filter only fires when the resolver runs (page render or REST call). Plugins calling `wp_admin_shell_register_menu_item()` from `init`+ are fine; calls from `plugins_loaded` or later via lazy callbacks may miss the current request if the resolver already ran and cached. Worth a one-liner in the docblock: "Call before the `init` action fires (or any earlier hook)." Same for `register_admin_route`.

tests/php/run-menu-route-shims-tests.php
```diff
@@ -0,0 +1,658 @@
...
+$ext_items = $filtered_ext['regions']['sidebar']['regions']['nav']['config']['items'];
+
+// existing item is at index 0; shim items follow
+WPAS_Shim_Test_Runner::assert_eq( 'https external href passes through', $ext_items[1]['href'], 'https://wordpress.org/' );
+WPAS_Shim_Test_Runner::assert_true( 'https flagged external', ! empty( $ext_items[1]['external'] ) );
+WPAS_Shim_Test_Runner::assert_eq( 'protocol-relative href passes through', $ext_items[2]['href'], '//cdn.example/' );
+WPAS_Shim_Test_Runner::assert_true( 'protocol-relative flagged external', ! empty( $ext_items[2]['external'] ) );
+WPAS_Shim_Test_Runner::assert_eq( 'hash href passes through unchanged', $ext_items[3]['href'], '#/inside-shell' );
+WPAS_Shim_Test_Runner::assert_true( 'hash NOT flagged external', empty( $ext_items[3]['external'] ) );
+
```
dabowman, line 433: Suggestion: lock down `external => false` semantics with an explicit test. Currently the impl auto-detects absolute URLs regardless of explicit-false; a test case would pin the contract (one way or the other) before plugin authors discover it.

wp-admin-shell.php
```diff
@@ -156,6 +158,58 @@ function wp_admin_shell_register_shell( $slug, $admin_json ) {
...
+ * `parent`, `parent_type`, `position`) carry over 1:1; the shell adds
+ * an optional `region` arg (defaults to the first `core:navigation`
+ * region in the resolved tree).
+ *
+ * Timing: call from `init` priority 9 or earlier (`plugins_loaded` is
+ * fine). The cascade resolver's first run on the page render or first
+ * REST hit triggers `wp_admin_shell_data_plugin` and memoizes the
+ * resolved tree through `WP_Admin_Shell_Cache`. Registrations made
+ * after the resolver's first run miss the current request entirely
+ * and only land on the next cache-invalidating event (option write,
```
dabowman: Timing note is right but understates the invalidation gap. `WP_Admin_Shell_Cache::key_for()` hashes shell slug + shells/*.json mtime + site/role/user option signals + user roles — registry state is NOT in the signal set. So registrations between page loads (e.g., plugin toggles a flag that changes `register_menu_item` calls) won't invalidate the cached tree even though the contribute output would differ. Steady-state deterministic registrations are fine (same hash → cached tree already contains them). Two options: (a) add a registry-fingerprint signal to `cache_key_for()` — `md5(serialize(WP_Admin_Shell_Menu_Items::all() + WP_Admin_Shell_Admin_Routes::all()))`, or (b) extend the docstring caveat: "registrations must be deterministic per page load; dynamic toggles need a manual `WP_Admin_Shell_Cache::flush()` to take effect."

dabowman (commented):
Re-review on 2e31f56 fix commit. All seven previous items addressed cleanly:

- `external => true|false` escape hatch landed (lines 332-340) + pinned with two new asserts.
- `array_merge` flipped to `static_data + config` so explicit `config` wins on collision (line 114) + collision test + extras-survive test.
- Docstring expanded (lines 46-65) covers screen-id constraint, parent heuristic, cross-region drop.
- Timing notes added to both public-API docblocks (init priority 9 caveat).
- Spec §13 #9 + #10 prose synced.
- CLAUDE.md test counts updated (915 → 919; menu-route file 57 → 61).

Four new items for awareness — none block.

includes/cascade/class-wp-admin-shell-menu-items.php
```diff
@@ -0,0 +1,509 @@
...
+		if ( ! empty( $item['icon'] ) ) {
+			$shell['icon'] = (string) $item['icon'];
+		}
+		if ( ! empty( $item['description'] ) ) {
+			$shell['description'] = (string) $item['description'];
+		}
+		if ( ! empty( $item['capability'] ) ) {
+			$shell['capability'] = (string) $item['capability'];
+		}
+		return $shell;
```
dabowman, line 372: `is_absolute_url` matches any RFC-3986 scheme — including `javascript:`. `to: 'javascript:alert(1)'` passes through `convert_to_to_href`, flows to React `<a href={...}>`. React 18 warns; React 19 still renders. Plugins are trusted (CIAB model), so out of strict scope, but worth a defensive allowlist: `preg_match('#^(https?|mailto|ftp)://#i', $value)` or explicit scheme check. Same regex used by `to_shell_link` for the `external` auto-flag. Low risk — flag for completeness.

includes/cascade/class-wp-admin-shell-menu-items.php
```diff
@@ -0,0 +1,509 @@
...
+	}
+
+	private static function to_shell_screen( $item, $kids ) {
+		$shell = array(
+			'screen' => (string) $item['id'],
+			'label'  => (string) $item['label'],
+			'items'  => array_values( $kids ),
+		);
+		if ( ! empty( $item['icon'] ) ) {
```
dabowman, line 363: Path-template `to` values pass through unchanged. CIAB plugins shipping `to: '/posts/{id}'` get the literal string `#/posts/{id}` as href — no current-context substitution. Probably matches CIAB's nav-level behavior (TanStack interpolates at navigation time, not menu render), but worth confirming + documenting. If interpolation is expected, the param source needs defining (route slot? entity context?).

includes/cascade/class-wp-admin-shell-menu-items.php
```diff
@@ -0,0 +1,509 @@
...
+
+	private static function to_shell_link( $item ) {
+		$shell = array(
+			'label' => (string) $item['label'],
+			'href'  => self::convert_to_to_href( $item['to'] ?? '' ),
+		);
+		if ( ! empty( $item['icon'] ) ) {
+			$shell['icon'] = (string) $item['icon'];
+		}
+		if ( array_key_exists( 'badge', $item ) && $item['badge'] !== null ) {
```
dabowman, line 338: `badge` passes through arbitrary value. Schema check? `array_key_exists('badge', $item) && $item['badge'] !== null` — `badge => ['html' => '<script>']`, `badge => new stdClass()`, `badge => false` all flow into the nav-item shape. The navigation app probably renders `<Badge>{item.badge}</Badge>`, which React string-coerces. Likely safe in practice. Tighten via `is_scalar($item['badge'])` if you want a contract floor.

includes/cascade/class-wp-admin-shell-cache.php
```diff
@@ -87,6 +95,7 @@ public static function key_for( $context = array() ) {
 			'user_prefs'  => self::user_meta_signal( 'wp_admin_shell_user_prefs' ),
 			'user_roles'  => self::current_user_roles(),
 		);
+		$signals = apply_filters( 'wp_admin_shell_cache_signals', $signals, $context );
```
dabowman, line 98: `wp_admin_shell_cache_signals` is now a public filter — any plugin can hook it to inject custom invalidation signals. Worth promoting from "in-process contributor coordination" to an explicit extension point in spec §13 (alongside the existing `wp_admin_shell_data_*` filters). Field-collections has the same static-registry-state cache gap and would benefit from hooking it the same way — out of C3 scope per the commit message, but tracking somewhere (feedback log?) keeps it from getting forgotten.

includes/cascade/class-wp-admin-shell-menu-items.php
```diff
@@ -0,0 +1,571 @@
...
+	 * rejected at register time so it never reaches a React `<a href>`.
+	 */
+	private static function is_safe_to( $to ) {
+		if ( ! is_string( $to ) || $to === '' ) {
+			return true;
+		}
+		if ( $to[0] === '/' || $to[0] === '#' ) {
+			return true;
+		}
+		if ( strpos( $to, '//' ) === 0 ) {
```
dabowman: Dead branch. Line 421 (`$to[0] === '/'`) already returns true for any string starting with `/`, including the `//cdn.example` protocol-relative form. Line 424 (`strpos( $to, '//' ) === 0`) never fires. Drop the block or keep with a `// Documented redundantly, in case the leading-slash check moves` comment. Cosmetic.

dabowman (commented):
Re-review on f36fd1a fix commit. All four prior items addressed:

- **Cache invalidation gap** → `wp_admin_shell_cache_signals` filter introduced in `class-wp-admin-shell-cache.php:99`; both shim files hook in with `md5( wp_json_encode( $registry ) )`. 7 new asserts pin baseline/after-register/after-delta/after-reset transitions.
- **`javascript:` / `data:` etc.** → `is_safe_to` allowlist at menu-items.php:417 (http(s)/ftp(s)/mailto/tel/sms + relative `/`/`#`/plain). 4 dangerous schemes rejected, 5 safe ones accepted in tests.
- **Path-template `to`** → docstring on `convert_to_to_href` explains router does the binding, not the shim.
- **`badge` type laxity** → `is_scalar` floor + 4 asserts.

Test count 61 → 81 matches the diff additions (9 + 4 + 7).

Two small items below — both nits.

dabowman (commented):
Round-4 fix `e6ea370` closes both nits cleanly:

- **`wp_admin_shell_cache_signals` documented** — spec §13 now has 11 entries; #11 spells out signature, default-signal coverage, recommended hook pattern. CLAUDE.md mirrors.
- **Dead branch dropped** — `is_safe_to` strips the redundant `strpos === 0` block; new comment on the leading-`/` check explains it covers both root-relative + protocol-relative.
- **Field-collections follow-up tracked** — logged to `docs/feedback.md` inbox with the 3-line patch sketch + ~5-assert test note. Won't get lost.

Nothing else to flag.

dabowman:
LGTM. Scope matches `docs/plans/track-b-menu-route-shims.md`; minor scope drift (cache-signals filter) is well-justified and documented as spec §13 #11. Tests green in the worktree (PHP suite not run per task scope).

## Verified against plan

| Plan item | Diff |
|---|---|
| Public `wp_admin_shell_register_menu_item()` + `wp_admin_shell_register_admin_route()` | ✅ `wp-admin-shell.php:1168` + `:1194` — free functions delegating to registry classes |
| Items land in resolved cascade tree | ✅ Both classes register `wp_admin_shell_data_plugin` callback at priority 5 (`admin-routes.php:271`, `menu-items.php:329`); mirrors `class-wp-admin-shell-field-collections.php:219` |
| Default nav-region resolution | ✅ `menu-items.php:511 find_default_nav_region_id()` walks `regions` tree for first `app === 'core:navigation'`; returns slash-path |
| `parent_type=drilldown` nesting | ✅ `build_tree()` (line 302) detects screen-parent EITHER via explicit `parent_type=drilldown` OR implicit (referenced by a child's `parent`); emits `to_shell_screen()` with nested `items[]` |
| `parent_type=dropdown` → drilldown fallback | ✅ `register()` line 479 coerces `parent_type` to `'drilldown'` after `warn_dropdown_fallback()`; per-id warn-once map |
| `gc_time` accepted + ignored | ✅ `admin-routes.php:207` calls `warn_gc_time()`; line 211 stores `$route` without `gc_time`; per-path warn-once map |
| Cap gating through 4 layers | ✅ `to_shell_link()` + `to_shell_screen()` copy `capability` onto the emitted item; `wp-admin-shell.php:539 wpas_collect_nav_item_caps()` recurses through `items[]` so the JS cap-pruner sees shim caps |
| Duplicate id rejection | ✅ `menu-items.php:417` + `admin-routes.php:170` return `WP_Error` with namespaced codes |
| End-to-end test (admin.json silent, resolved tree contains shim entries) | ✅ `run-menu-route-shims-tests.php:961-998` exercises `apply_filters('wp_admin_shell_data_plugin', $bare)` |
| Spec §13 #10 + #11 + (bonus #11) | ✅ `docs/wp-admin-shell-design-spec.md:891` "Eleven extension points"; #9 menu / #10 route / #11 cache-signals all present |
| CLAUDE.md updated | ✅ Test count `843 → 939`; extension-points block extended |
| Variant-addressable nav items | ✅ OUT per plan; explicitly called out in spec §13 #9 trailing note |
| Per-item deregistration API | ✅ OUT per plan; `register()` returns `WP_Error` on dup, no `unregister()` |

## Tests (worktree)

```
lint:js          clean
test:schema      79 / 79
test:runtime    14 / 14  (final file output; full suite chained)
build           clean (3 webpack perf warnings, pre-existing)
```

PHP suite (`run-menu-route-shims-tests.php`) not executed per task scope. Asserted-count audit: 74 static `assert_*` calls + two `foreach` loops over 4-bad-scheme + 5-ok-scheme arrays = **81 assertions**, matches the CLAUDE.md claim. PR body's "57/57" stat is stale from an earlier round of the branch.

## Scope drift

Plan listed 5 file paths; PR touches 8:
- `includes/cascade/class-wp-admin-shell-cache.php` (+9 lines) — adds `apply_filters('wp_admin_shell_cache_signals', ...)` to `key_for()`. **Not in plan, but defensible**: without it the static-class registries are invisible to the cache key and a registration delta between requests serves stale resolver tree. Each new registry hooks the filter at module load (`admin-routes.php:276`, `menu-items.php:334`). Spec §13 grew to ELEVEN points; plan budgeted for ten.
- `docs/feedback.md` (+1 line) — files a `[2026-05-14][bug]` noting `WP_Admin_Shell_Field_Collections` has the same cache-invalidation gap and needs the same 3-line hook. Correct triage; out-of-scope for C3.

Recommend resolving Track A coordination via rebase: if Track A merges first with its own #9 (REST preload), this PR's entries become #10 / #11 / #12, and the spec prose "Eleven" → "Twelve". `CLAUDE.md` test-count line rebases the same way.

## Code review notes

1. **Cross-region parent silently demotes child to root.** `menu-items.php:299` groups items into `$by_region` BEFORE `build_tree()` runs; a child whose `parent` lives in a different region bucket falls out of `$id_set` inside `build_tree()` and emits as a root link instead of nesting. Documented in the class-level comment (line 56–58) but no test covers the case. Plugin authors will hit this if they refactor a single drilldown into "parent in default region, children in named region" — file a follow-up test asserting the documented behavior, or strengthen to a `WP_DEBUG` notice on cross-region orphan.

2. **`is_safe_to` allowlist is tight.** `menu-items.php:732` allowlists `http(s)`, `ftp(s)`, `mailto`, `tel`, `sms` + relative paths + protocol-relative; `is_absolute_url` correctly skips parsing relative paths. `javascript:` / `data:` / `file:` / `vbscript:` all rejected at register time (test verifies). Defense-in-depth is the right call even with trusted plugin code.

3. **Empty-registry no-op is properly guarded.** `contribute()` early-returns on empty registry (line 240 / 552) — no false-positive routes/items appended on a bare admin.json.

4. **`wpas_collect_nav_item_caps` recursion covers shim screens.** `to_shell_screen()` line 672 preserves `items[]`; cap collector at `wp-admin-shell.php:539` recurses through them, so a `capability` on a child of a drilldown screen registers in the JS-side cap payload. Verified via the cap-collector test pair (line 1044–1054).

5. **Sort stability.** `sorted_items()` line 593 wraps each item with its registration index `$i` and tie-breaks on it after `position` compare. Null position → `PHP_INT_MAX` → registration order preserved within nulls. Correct.

6. **`config` merge direction is documented and tested.** `admin-routes.php:201` does `array_merge( $static_data, $config )` so explicit `config` wins on collision; test at line 877–893 covers both the merge direction and that no-collision static_data extras survive. Matches CIAB's "static_data as preloaded base" semantics.

7. **Schema-pattern parity.** `admin-routes.php:129` literal-copies the `routes` pattern from admin-v2 (`^/[A-Za-z0-9_/{}\-*]*$`). When that pattern moves, both copies need a touch — worth a comment cross-link, but no live drift today.

## Gates remaining

- [ ] **Run PHP suite via `wp eval-file run-menu-route-shims-tests.php`** — expect 81 / 81.
- [ ] **Manual smoke via test plugin**: register one drilldown screen + two child items + one admin-route through the shims; load `developer-admin.json`; verify nav appears in sidebar drill-down + route resolves.
- [ ] **Rebase against Track A** once it lands (#9 numbering + CLAUDE.md test-count line).
- [ ] **Follow-up commit on C2 surface**: port the `wp_admin_shell_cache_signals` hook to `WP_Admin_Shell_Field_Collections` per the feedback inbox note — the gap is real, the fix is the same three lines.

PR #44, opened 2026-05-15
F.2: migrate core:users to useViewConfig

dabowman:
## Summary

- Track F.2 of the entity-CRUD migration sweep (`docs/plans/track-f-entity-crud-migrations.md`).
- UsersApp now consumes `useViewConfig('root', 'user')`; structural DataViews config (fields / actions / defaultView / defaultLayouts) moves to `app.json#viewConfig` as the manifest baseline. Admin.json `viewConfigs.root.user._default` + filter `wp_admin_shell_view_config_root_user` override the cascade.
- Parity-preserving — same self-delete guard, same bulk delete-with-reassign modal, same role filter. Only externally visible change is that admin.json can retune the table without a code change.
- Adopts post-E PostsApp template verbatim: LABELS tables (i18n recovery), `buildFields` / `buildActions` / `buildFieldRenderers` keyed by spec id, view-state resync `useEffect`, title-field dedup memo.
- Switched modal `Promise.all` → `Promise.allSettled` so one rejection doesn't collapse a bulk delete; partial-failure surfaces via a count-aware error notice.
- `developer-admin.json` gets a `viewConfigs.root.user._default` override (`perPage: 10`) to exercise the cascade end-to-end in the bundled demo.
- `app.json#documentation.data.reads` lists the kernel-config view-config triple first.
- `app.md` grows an Architecture section + Translation recipe + Known limitations note on override semantics.
- CLAUDE.md app table flags `core:posts` + `core:users` as view-config consumers.

## Test plan

- [x] `npm run test:schema` — 90 / 90 (bundled `core:users` manifest validates against `admin-app-v2.json`; bundled `developer-admin.json` validates against `admin-v2.json` w/ the new `root.user._default` override)
- [x] `npm run test:runtime` — full sweep clean
- [x] `npm run test:parity` — 4 / 4
- [x] `npm run build` — clean (size warnings unchanged; same vendors chunk)
- [ ] PHP suites (`run-view-config-tests.php` etc.) — no PHP changed; suites remain green from main. Re-run locally before merge if wp-env is up
- [ ] Browser smoke (manual, on shell with users route):
  - [ ] `/users` loads; table renders name + email + roles + registered columns
  - [ ] Sort by name / registered date works
  - [ ] Search filters rows
  - [ ] Role filter (single-select `is`) narrows rows
  - [ ] Delete on a single non-self user — confirm modal, success snackbar, row gone after invalidate
  - [ ] Delete on a bulk selection including self — modal copy notes "(Your own account will be skipped.)", destructive proceeds for the others
  - [ ] Delete on a bulk of self-only — destructive button disabled
  - [ ] Locale switch (e.g. `de_DE`) — column / action labels translate via `FIELD_LABELS` / `ACTION_LABELS`
  - [ ] Cascade override: `developer-admin.json` lands `perPage: 10`; the table shows 10 rows per page (vs manifest baseline 20). Confirm reload picks up override and `wp_admin_shell_view_config_root_user` filter wins last if registered

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -296,13 +296,13 @@ wp-admin-shell/
 
 | Source | Component | Native? | Cap floor | Notes |
 |---|---|---|---|---|
-| `core:posts` | PostsApp | ✅ | — | DataViews table; `config.postType` |
+| `core:posts` | PostsApp | ✅ | — | DataViews table; `config.postType`; consumes view-config `(postType, {postType}[, variant])` |
 | `core:simple-editor` | SimpleEditorApp | ✅ | — | Substack-style; title + 9 blocks + auto-save |
 | `core:editor` | EditorApp | iframe | — | `post.php?post={id}&action=edit`. Native `@wordpress/edit-post` mount deferred to v2.x — see `SiteEditorApp.js` for blockers. |
 | `core:media` | MediaApp | ✅ | — | Grid, upload, detail modal |
 | `core:taxonomy` | TaxonomyApp | ✅ | — | DataViews + create/edit/delete terms |
 | `core:profile` | ProfileApp | ✅ | — | `useEntityRecord('root','user',userId)` |
-| `core:users` | UsersApp | ✅ | `list_users` | DataViews + bulk delete with reassign + self-delete guard |
+| `core:users` | UsersApp | ✅ | `list_users` | DataViews + bulk delete with reassign + self-delete guard; consumes view-config `(root, user)` |
 | `core:comments` | CommentsApp | ✅ | `moderate_comments` | DataViews + approve/spam/trash via partial saveEntityRecord |
 | `core:settings` | SettingsApp | partial | `manage_options` | Composable host; native general/writing/reading/discussion + iframed permalinks/media/privacy |
 | `core:settings-general` | SettingsGeneralApp | ✅ | — | Standalone version of the General panel (legacy entry; kept registered) |
--- a/shells/developer-admin.json
+++ b/shells/developer-admin.json
@@ -359,6 +359,30 @@
 					]
 				}
 			}
+		},
+		"root": {
+			"user": {
+				"_default": {
+					"fields": [
+						{ "id": "name", "type": "text", "label": "Name", "enableGlobalSearch": true, "enableHiding": false },
+						{ "id": "email", "type": "text", "label": "Email", "enableGlobalSearch": true },
+						{ "id": "roles", "type": "text", "label": "Roles", "filterBy": { "operators": [ "is" ] } },
+						{ "id": "registered_date", "type": "datetime", "label": "Registered" }
+					],
+					"defaultView": {
+						"type": "table",
+						"page": 1,
+						"perPage": 10,
+						"fields": [ "email", "roles", "registered_date" ],
+						"titleField": "name",
+						"sort": { "field": "name", "direction": "asc" }
+					},
+					"defaultLayouts": { "table": {}, "grid": {} },
+					"actions": [
+						{ "id": "delete", "label": "Delete", "isDestructive": true, "supportsBulk": true, "icon": "trash" }
+					]
+				}
+			}
 		}
 	},
 
--- a/src/apps/users/app.json
+++ b/src/apps/users/app.json
@@ -14,6 +14,31 @@
 		"additionalProperties": false
 	},
 	"script": "wp-admin-shell",
+	"viewConfig": {
+		"kind": "root",
+		"name": "user",
+		"defaultView": {
+			"type": "table",
+			"search": "",
+			"filters": [],
+			"page": 1,
+			"perPage": 20,
+			"sort": { "field": "name", "direction": "asc" },
+			"fields": [ "email", "roles", "registered_date" ],
+			"titleField": "name",
+			"layout": {}
+		},
+		"defaultLayouts": { "table": {}, "grid": {} },
+		"fields": [
+			{ "id": "name", "type": "text", "label": "Name", "enableGlobalSearch": true, "enableHiding": false },
+			{ "id": "email", "type": "text", "label": "Email", "enableGlobalSearch": true },
+			{ "id": "roles", "type": "text", "label": "Roles", "filterBy": { "operators": [ "is" ] } },
+			{ "id": "registered_date", "type": "datetime", "label": "Registered" }
+		],
+		"actions": [
+			{ "id": "delete", "label": "Delete", "isDestructive": true, "supportsBulk": true, "icon": "trash" }
+		]
+	},
 	"window": {
 		"defaultSize": {
 			"w": 1024,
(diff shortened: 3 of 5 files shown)


dabowman:
LGTM. Parity-preserving migration, every box in the shared F-scope checklist ticked; tests green in the worktree.

## Verified against plan

| Plan item (F shared scope) | Diff |
|---|---|
| `app.json#viewConfig` block (kind/name/fields/actions/defaultView/defaultLayouts) | `src/apps/users/app.json:17-41` — `kind:"root", name:"user"`, four fields, single `delete` action, `titleField:"name"` |
| `app.json#documentation.data.reads` lists kernel-config first | `src/apps/users/app.json:60-70` — `viewConfigs.root.user._default` precedes `root/user` core-data read |
| `index.js` imports `useViewConfig` + drops inline state seeds | `src/apps/users/index.js:10,316` |
| `FIELD_LABELS` + `ACTION_LABELS` tables w/ `?? spec.label` precedence | `src/apps/users/index.js:18-27, 110, 274` |
| View-state resync `useEffect` | `src/apps/users/index.js:329-335` (see note 1) |
| Title-dedup pattern (`titleField` stripped from `view.fields`) | `src/apps/users/index.js:389-400` — memoized `visibleView`; matches PostsApp recipe |
| `app.md` Architecture + Translation recipe + parity-gap | `src/apps/users/app.md:13-50, 76` (Roles `elements` note) |
| Cascade override in `developer-admin.json` | `shells/developer-admin.json:362-383` — `perPage:10` override |
| CLAUDE.md app table flags view-config consumers | `CLAUDE.md:299,305` — `core:posts` + `core:users` |
| F.2 divergence: bulk delete-with-reassign modal stays in `index.js` | `src/apps/users/index.js:118-261` |
| F.2 divergence: self-delete guard preserved | `src/apps/users/index.js:124-127` — `targets = items.filter( i => i.id !== currentUserId )` |
| F.2 divergence: `(root, user)` entity triple | `src/apps/users/index.js:316` — `useViewConfig('root', 'user')` |

## Tests (worktree)

```
lint:js         clean
test:schema     90 / 90
test:runtime   17 / 17 (final block; full chain clean)
test:parity     4 / 4
build          clean (251 KiB entrypoint warning unchanged from main)
```

PHP suites untouched by the diff — skipped re-run.

## Code review notes

1. **View-resync `useEffect` deps are `[]`, not `[axis]`.** `src/apps/users/index.js:329-335` runs once on mount with the same shape as the `useState` initializer at line 318 — effectively a no-op. PostsApp keys on `[postType, variant]` because the same hook instance can flip triples; UsersApp has no variant axis today and `useViewConfig('root', 'user')` is constant, so there's nothing to resync on. The inline comment is honest about this ("UsersApp ships a single triple today, but the recipe matches PostsApp's so a future variant config ... picks up without rewriting"). Fine to ship as dead-code-but-future-proof; if a `role` variant axis ever lands, deps need to become `[role]` or the equivalent. Non-blocking.
2. **`Promise.all` → `Promise.allSettled` upgrade is a genuine improvement.** `src/apps/users/index.js:193-250` — three buckets (zero / partial / all-failed) with distinct UX. Plural snackbar for full success, dismissible error notice with failed/total counts for partial, fallback error notice with first rejection's message for all-failed. `_n()` plural form is correctly keyed on `targets.length` (total count for grammatical agreement), matching the WP i18n convention.
3. **`onActionPerformed?.( targets )` moved out of the try/catch.** `src/apps/users/index.js:251` — now always fires after settle, whereas pre-PR it only fired on success. Behaviorally desirable: DataViews needs to clear selection regardless of partial-failure state. Verified intentional vs. the pre-PR behavior.
4. **`invalidateResolution` runs unconditionally after settle.** `src/apps/users/index.js:211-214` — correct; the cache is stale either way once any delete succeeded, and an extra refetch on all-failed is harmless.
5. **`buildFieldRenderers()` invoked fresh in every `useMemo`.** `src/apps/users/index.js:403` — re-creates the render closures on every `viewConfig` change. Renderers are stable (no closed-over state), so this is wasted-work but not a correctness bug. Same shape PostsApp ships (`src/apps/posts/index.js:362`). Skip.
6. **Roles renderer assumes `item.roles` is already joined.** `src/apps/users/index.js:67` — works because the data-mapping memo does `( record.roles || [] ).join( ', ' )` at line 381. Coupling lives in the same component, low risk.
7. **`developer-admin.json` cascade override carries no `titleField`.** `shells/developer-admin.json:362-383` — `defaultView` lists `email, roles, registered_date` but omits `titleField`. The visibleView memo falls back to `viewConfig.defaultView?.titleField` (`src/apps/users/index.js:391`) — but since admin.json wins as whole-entry override, this is the *post-merge* `viewConfig.defaultView`, which is the admin.json shape (no titleField). Net effect: `name` is omitted from `view.fields` (good — the admin.json author wanted email-first) AND `view.titleField` is undefined (DataViews renders no title column). Whether that's intended depends on what the cascade demo is meant to show. Flag for the manual smoke: confirm the demo shell renders the name column at all. Could fix by adding `"titleField": "name"` to the override.
8. **`fieldsRef` listed in app.json#documentation.data.reads** (`app.json:68`) but not consumed by `index.js`. Fine — that's documentation of what the kernel-config triple *can* carry, not what the app personally reads. Matches PostsApp.
9. **No scope drift.** Five files touched; all in plan. No conflict surface against C/D/E.

## Gates remaining (PR test-plan items still unchecked)

- [ ] Browser smoke: `/users` table renders four columns; sort by name/registered date; search; role filter (single `is` operator).
- [ ] Bulk delete with self in selection — confirm the "(Your own account will be skipped.)" addendum + partial flow.
- [ ] Bulk delete with self-only — destructive button disabled, error notice on click.
- [ ] Locale switch (`de_DE`) — column + action labels translate via `FIELD_LABELS` / `ACTION_LABELS`.
- [ ] Cascade override end-to-end: `developer-admin.json` lands `perPage:10`; confirm + verify whether the missing-`titleField` in the override is intended (see note 7).

To run locale smoke:

```bash
npx wp-env run cli wp language core install de_DE
npx wp-env run cli wp option update WPLANG de_DE
# Reload shell; navigate /users; confirm column headers translate.
```

After the gates pass (especially note 7's `titleField` check on the demo shell), squash-merge — F.2 is mutually independent of F.1/F.3/F.4/F.5 and unblocks the next sub-track agents.

dabowman:
Re: review notes —

**Note 7 (`developer-admin.json` cascade override missing `titleField`):** double-checked — the override already carries `"titleField": "name"` at `shells/developer-admin.json:377`. Re-reading the block:

```json
"defaultView": {
	"type": "table",
	"page": 1,
	"perPage": 10,
	"fields": [ "email", "roles", "registered_date" ],
	"titleField": "name",
	"sort": { "field": "name", "direction": "asc" }
},
```

No JSON edit needed. The cascade smoke remains: reload with this shell active → expect 10 rows/page (vs manifest baseline 20) and name as the title column. Will verify in the manual browser smoke gate.

**Notes 1–6, 8, 9** — acknowledged; non-blocking, no diff. The view-resync `useEffect` (note 1) stays as the recipe-consistent stub; will tighten the deps if a variant axis (e.g. `?role=`) ever lands.

**Smoke gates** — still pending against a live wp-env. Will run them before merging, including the `de_DE` locale check + the `perPage:10` cascade override verification.

Issue #1, opened 2026-04-30
Harden selection-bus REST endpoint


dabowman:
Group of small REST hardening items found in M1 review.

### 1. Unbounded payload size (medium)
`includes/class-wp-admin-shell-selection-rest.php:78–88` — `set_one()` writes `$value` straight to user_meta with no size cap or shape validation. Authenticated user can stuff arbitrary blobs into own user_meta via repeated POSTs. Per-user gated so blast radius is self-DoS only, but cap (~64KB) before M5 prod readiness pass.

### 2. `permission_check` is bare `is_user_logged_in()`
`includes/class-wp-admin-shell-selection-rest.php:62` — selection writes hit own-user meta so isolation holds, but flag for M5 four-layer cap-gating pass.

### 3. Scope regex excludes `:`
`includes/class-wp-admin-shell-selection-rest.php:38` — route pattern `[A-Za-z0-9._-]+` blocks future namespaced scopes (e.g. `posts:selection`). Loosen when M2 needs it.

### 4. Silent error swallowing in `persist.js`
`src/runtime/selection/persist.js:44/61/75` — silent catch is intentional + commented. Optional `console.warn` behind dev flag would aid debugging. Not a behavioural bug, but worth a dev-mode log.

Source: M1 review of commits `99a8627` + `8321f73`. Logged in `docs/feedback.md`.

dabowman:
Fixed in 2fb45f4198cd33c37f18274f73c90664e52f3395. 64 KiB cap + regex loosening + dev warn + permission-check intent comment. Verified.

PR #49, opened 2026-05-20
Reshape admin.json schema around user-task surfaces (v3)

dabowman:
## Summary

Reshape admin.json from a runtime-pipeline-mirror into a user-task-surface schema. Collapses the five high-cost authoring intents (add a screen, add a column, reorganize menu, restrict by capability, replace a screen) to single-block edits each, and adopts theme.json patterns where they fit.

**State:** feature-complete and hardened. All planned phases (3a → 3d) plus the full post-review hardening pass have landed. v3 default workspace renders end-to-end; the v2 default shell has been retired and all six demo shells migrated to v3. Targeting **v3.0.0-beta.1**.

## Top-level shape

```
workspace   install metadata — engine, default-screen, branding, notices, persistent widgets
settings    registries — dataViews (3-axis: kind → name → variant), dataFields
screens     id-keyed map; each entry bundles route + app + view + permissions + mode
menu        nested tree (item key matching a screen id implicitly binds label/icon/permissions)
commands    first-class shortcuts + palette entries with explicit id (replaces v2 bindings)
permissions OR-semantic — capabilities[] + roles[], trust-tier cascade, app-floor AND backstop
modes       engine-declared chrome catalog with `extends` inheritance; visibility is paint-only
slots       three tiers (kernel + engine + app), two scopes (workspace + screen)
styles      theme.json-shaped
preload     REST hydration
regions / routes  escape hatches
```

Naming: "shell" (user-facing concept) → **workspace**. Plugin + PHP class names keep "WP Admin Shell" (runtime). `viewConfig` → `dataView`; `viewConfigs` → `settings.dataViews`; `fieldCollections` → `settings.dataFields`; `default-route` → `workspace.default-screen`; `bindings` → `commands`.

## What landed

**Schema family** — `docs/schemas/admin-v3.json` + `admin-app-v3.json` + `admin-engine-v3.json`, with positive/negative fixtures.

**Cascade + compiler**
- Null-tombstone support in `WP_Admin_Shell_Merge` at any depth (theme.json convention), gated behind trust-tier authority.
- v3 compiler (`class-wp-admin-shell-v3-compiler.php`) — detects v3 shape, synthesizes `routes`/`regions`/`default-route`/`commands` from the v3 blocks; v2 shells pass through. Multi-app slot-keyed route synthesis.
- 3-axis dataView resolver (`(kind, name, variant)` with `extends` chains, cycle/depth caps, ref-wins-inline merge); `settings.dataFields` collections.
- Permissions resolver — OR-semantic eval, trust-tier enforcement + audit log, super-admin magic, fail-closed unknowns.
- Mode resolution — `extends` chain with circular-ref detection, plugin-contributed modes via filter, modal LIFO stack.
- Engine `defaultRegions` for core:default / core:single-pane / core:desktop.

**Runtime**
- `useDataView` (renamed from `useViewConfig`/`useScreenView`) — 3-axis + screen-keyed lookup sharing one LRU-capped cache; per-render cancellation + inflight dedup.
- Navigation app consumes the nested `menu` tree; drilldown state in URL slot `?screen=`.
- `BindingsConsumer` reads `commands` (memoized compile); `userCan` rewritten for OR semantics (`userCanAccessScreen`).
- Region.js applies mode region-state as `data-mode-*` attributes + BEM classes (display:none, mount tree stable).
- Classic wp-admin menu bridge, dashboard-host rewrite (`apps[]` + `slot:"grid"`), command-palette rewrite (`commands[]` + screen-derived entries).

**Workspaces** — `wp-admin-default-v3.json` + six migrated demo shells; `wp admin-shell migrate-shell <slug>` CLI for v2 → v3.

## Phase plan — all shipped
- ✅ 3a — schemas + cascade null-tombstone
- ✅ 3b — resolvers (dataView / menu / permissions / modes)
- ✅ 3c — dashboard-host rewrite (#53), command palette (#51), classic wp-admin menu bridge (#54), multi-app layout (#55)
- ✅ 3d.0 — dataView 3-axis registry restoration (#50)
- ✅ 3d.1 — migrate 6 shells + retire v2 default (#56)
- ✅ 3d.2 — v2 → v3 migration CLI (#58)
- ✅ 3d.3 — test surface rewrites (#60)
- ✅ 3d.4 — documentation sweep (#59)
- ✅ 3d.5 — pre-v3.1 shim-removal hardening (#57)

## Post-review hardening (all review blockers closed)

Multi-pass review surfaced 7 blockers + ~12 majors, tracked in `docs/plans/2026-05-22-pr49-pre-merge-feedback.md` as five sequenced PRs — all merged:
- **Security cascade (#61)** — trust-tier enforcement wired into the resolver (was dead code); null-tombstones gated behind `$authoritative`; per-field `customizable` enforcement restored with a hardcoded deny-list (`screens.*.permissions`, `screens.*.app`, `commands.*.invoke`, `workspace.engine`); `is_safe_href` rejects protocol-relative / whitespace-leading / backslash variants; `/data-view` REST gated per requested screen's permission floor.
- **`user-switchable` key fix** — code now reads the schema-canonical kebab key (was always-false in the v3 default install).
- **Runtime perf (#62)** — `useDataView` LRU cap; `BindingsConsumer` memoization; `useMode` hoisted off unstable refs.
- **Kernel DS-neutrality (#63)** — `WpdsThemeProvider` + `SlotFillProvider` relocated into engines; `data-wpds-theme-provider-id` → `data-theme-scope-id`; static-analysis neutrality guard added.
- **Hygiene (#62)** — schema `$id` numbering reconciled; filter priorities pinned apart; filter-vs-compiler ordering documented.

## Test surface (~2055 assertions)

| Side | Count | Verified this pass |
|---|---|---|
| Node — parity | 4 | ✅ 4/4 |
| Node — schema (admin v1/v2/v3 + app/engine/tokens sweeps) | 103 | ✅ 103/103 |
| Node — runtime (23 files) | ~917 | ✅ 0 failed |
| Node — engines (core:desktop, TS strip-types) | 77 | ✅ 77/77 |
| PHP — 18 runners (cascade / permissions / modes / v3-compiler / security-cascade / data-view / migrate-cli / classic-menu-bridge / …) | ~954 | ⚠️ see note |
| `lint:js` | — | ✅ clean |

**Verification note:** Node suites + lint were run green locally on `bc7fa9e`, and a `php -l` sweep over all 49 PHP files is syntax-clean. The PHP behavioral suites run via `wp eval-file` and require a provisioned WordPress (`wp-env`); they were **not** executed in the review container (the environment's network policy blocks the WordPress image download). They should be run green via `npx wp-env run cli wp eval-file …` before tagging.

## Remaining before tag (process, not features)
- [ ] Run the 18 PHP suites green in a wp-env-capable environment.
- [ ] Manual smoke per `docs/v1-readiness.md`: cap-gating across roles on `wp-admin-default-v3`, each migrated demo shell, Cmd+K palette, shell switching, form-save (PHP 8.1+).

## Deferred (post-beta, non-blocking)
- DataForm adoption (zero adopters today; taxonomy/media are the cheap wins).
- ~650 LOC dataView-utility dedup across the six entity-CRUD apps (`compileEligibility` / `composeFields` / `composeActions` / resync hook).
- v3.1 shim removals (v2-name filter/hook deprecation shims — informational now, load-bearing at v3.1).

## Design docs
- `docs/v3/schema-sketch.md` — locked design decisions.
- `docs/v3/roadmap.md` — phase status + open/deferred decisions.
- `docs/v3/core-default-engine.v3.md` — engine.json v3 shape with mode catalog.
- `docs/plans/2026-05-22-pr49-pre-merge-feedback.md` — the five-stage hardening plan.


dabowman:
## Code Review — v3 branch

### Overview
New v3 schema family (`admin-v3.json`, `admin-app-v3.json`, `admin-engine-v3.json`) + compiler (v2→v3 normalize), new cascade subsystems (`permissions`, `modes`), heavy menu-items rewrite, view-config rewrite for screen-keyed lookups, `useScreenView` replacing `useViewConfig`, navigation drilldown rework, mode-aware regions. Default v3 shell (`wp-admin-default-v3.json`, 1.1k lines). 2 new test files (mode-resolution + v3 compiler).

---

### 🔴 Blockers

- **Trust-tier permission enforcement is dead code.** `WP_Admin_Shell_Permissions::enforce_trust_tiers()` at `includes/cascade/class-wp-admin-shell-permissions.php:149` is never called. Combined with `class-wp-admin-shell-customizable.php:113-117` blanket pass-through, a `user`-origin doc can rewrite `screens.<id>.permissions.capabilities` and grant itself any cap. Wire into `Resolver::resolve_with()` per-origin or as a `wp_admin_shell_data_{origin}` filter. Add tests covering: user origin adds cap → rejected; site origin removes cap → applied.

- **Null-tombstone bypasses restrict-only enforcement.** `class-wp-admin-shell-merge.php:108-111` (assoc `null`) and `:205-212` (`__tombstone: true` in keyed arrays) fire regardless of `\$authoritative`. A user-origin doc sending `{ "screens": { "users": null } }` deletes trusted screens; same for `{ "menu": [ { "id": "users", "__tombstone": true } ] }`. v2's restrict-only invariant is broken. Gate both tombstone paths behind `\$authoritative === true`.

- **`customizable` enforcement removed for every v3 top-level block.** `class-wp-admin-shell-customizable.php:113-117` passes downstream `menu`/`screens`/`commands`/`workspace`/`preload`/`regions`/`routes` through verbatim. The comment promises enforcement \"lives elsewhere\" but it doesn't — see Blocker #1. At minimum, `screens[id].permissions`, `screens[id].app`, `commands[].invoke`, `workspace.engine` must be default-deny at consumer origins.

- **`useScreenView` stale-closure overwrite.** `src/runtime/viewConfig/useScreenView.js:87-121` — in-flight cache keyed by `usableId` only. If screen A's fetch resolves after user switched to B, A's `cancelled` flag was per-effect and is already past its check; the `setDoc` call clobbers B's already-rendered state. Capture `usableId` at effect entry in a `useRef` token; bail if mismatch.

- **`useScreenView` module-level cache unbounded.** `src/runtime/viewConfig/useScreenView.js:11` — every distinct `screenId` ever requested stays in memory for page life. Add an LRU cap (Map preserves insertion order; `delete` oldest at ~64).

---

### 🟠 Major

- **`is_safe_href` allows protocol-relative bypass.** `includes/cascade/class-wp-admin-shell-menu-items.php:480-502` final `return strpos( \$href, ':' ) === false;` lets `//evil.example.com/path` through (no colon). Add `if ( strpos( \$href, '//' ) === 0 ) return false;` before the trailing return.

- **`shell_allows_user_switch` reads the wrong key.** `class-wp-admin-shell-resolver.php:307` checks `\$doc['userSwitchable']` (camelCase) but both v2 and v3 schemas declare `user-switchable` (kebab). v2 latent; v3 inherits. User-origin shell selection silently never works.

- **v3 schemas `\$id` collides with v2 numbering.** v2 schemas use `\$id: https://schemas.wp.org/admin/v1.json` (literal `v1`), v3 uses `v3.json`, no `v2.json` `\$id` exists. A validator resolving `\$id` against `schemas.wp.org/admin/v1.json` fetches a v2-shape document. Fix v2 `\$id`s to say `v2.json` or document the discontinuity.

- **View-config `fieldsRef` lookup is unsanitized.** `class-wp-admin-shell-view-config.php:142-143` reads `\$config['settings']['fields'][ \$ref ]` directly. Schema regex runs at validation time but runtime-registered docs may skip. Run `WP_Admin_Shell_Field_Collections::sanitize_segment()` (or slash-aware variant) before lookup.

- **`useMode` recompute thrash.** `src/runtime/modes/useMode.js:58-83` builds `engineModes`/`screens` from referentially-unstable `config` each render and passes them as `useMemo` deps → memo busts every render. Compounded: `src/runtime/regions/Region.js:157` calls `useMode()` **per region**. Windowed engine = N child regions × per re-render × full mode resolve. Hoist mode resolution to a kernel-level context provider; read via `useContext`.

- **`BindingsConsumer` rebinds document keydown every render.** `src/runtime/bindings/BindingsConsumer.js:34-40, 86` — `commands` array built fresh outside `useEffect`, used as a dep. Every router event tears down + rebinds. Wrap construction in `useMemo([ config?.commands, config?.bindings ])`.

- **`parseShortcut` Alt-fallback over-matches.** `src/runtime/bindings/parseShortcut.mjs:107-121` falls back to `event.code === 'KeyN'` for any single-letter shortcut regardless of `needAlt`. Dvorak `Cmd+K` with a remapped physical key would spuriously match. Gate: `if ( needAlt && /^[a-z]\$/.test(keyLower) ) { … }`.

- **`useScreenView` snapshot read non-reactive.** `useScreenView.js:33` reads `window.wpAdminShell.config` once via `useMemo`; shell-switch / option-write leaves it stale. Use `useKernel().config` (same data, reactive).

- **`useMode` reads `window.wpAdminShell.engineModes` in render body.** `useMode.js:64-66` — not subscribed. Late-registered shell / hot reload doesn't trigger re-render. Freeze at boot + document, or surface through kernel-context.

- **v3-compiler slot-routed key collision silent.** `class-wp-admin-shell-v3-compiler.php:294` (`'@' . \$slot . '/' . \$screen_id`) — two screens sharing `slot=palette` + identical `path` → second loses to first via `isset( \$routes[\$slot_key] )` short-circuit. Emit `_doing_it_wrong` notice (matches v2 path-collision contract).

- **`Modes::deep_merge` cycle stamp loss.** `src/runtime/modes/resolveMode.mjs` — two parents in `extends` chain hitting depth-limit overwrite each other's `_extendsChainError`.

- **`merge_keyed_arrays` Pass 3** (line 252-258) preserves non-keyed entries from `over` without merging; stack of `{ separator: true }` entries accumulates across origins with no tombstone path. Auto-key separators or stamp origin.

---

### 🟡 Minor

- **Compiler runs after `wp_admin_shell_data` filter.** `class-wp-admin-shell-resolver.php:122-134` — plugin-injected `screens`/`menu`/`commands` invisible to `inject_app_baselines` (priority 5, before compile). Run compiler before filter or document the constraint.
- **`forward_v2_bindings_to_commands` ids include `\$i`.** `class-wp-admin-shell-v3-compiler.php:205` — cascade-driven position shift mutates synth id; user-origin overrides break silently. Drop `\$i`.
- **`is_known_capability` only walks current user's caps.** `class-wp-admin-shell-permissions.php:456-459` — CLI/snapshot for arbitrary user fail-closes legitimate caps. Take `\$user_id` parameter or document.
- **Priority collision on `wp_admin_shell_data`.** `class-wp-admin-shell-menu-items.php:528` (`bind_screens`) and `class-wp-admin-shell-view-config.php:483` (`inject_app_baselines`) both at priority 5; WP filter order undefined. Pin apart.
- **Compiler `is_v3` detection over-broad.** Any top-level `workspace` block flips v3 detection. Bump to `version === 3`.
- **`Region.js` state class collision.** Region id `hidden` vs state key `hidden` → identical class. Suggest `wp-admin-shell-region--state-\${key}`.
- **`useScreenView` double-effect flicker.** Two effects same deps → `doc → null → fetched` flicker. Combine into one effect.
- **`useScreenView` error swallowed.** `.catch(() => …)` silently drops REST error. Add `console.warn` gated on dev build.
- **`findScreen` only top-level.** `src/apps/navigation/index.js:496-503` — 3-level menus break silently. Spec drilldown is 2-level today but latent.
- **No automated tests** for `WP_Admin_Shell_Permissions`, v3 menu-items path, or null-tombstone cascade semantics. Add before tagging.

---

### ⚪ Nits

- v3 schema `menuItem` pattern (line 152, 410) allows `:` after first id segment; validator at `menu-items.php:90-94` only checks non-empty string. Drift.
- `menu-items.php:514` `trigger_error( esc_html( \$message ), E_USER_NOTICE )` — `esc_html` on `WP_DEBUG` trace is dead. Switch to `error_log` (matches `permissions.php:553`).
- `synthesize_default_route` (`v3-compiler.php:396`) iteration order ≠ lexical. Sort `array_keys( \$screens )` first.
- `useScreenView` `fallbackDoc = fallback || {}` (line 68) creates fresh `{}` per render. Hoist `const EMPTY = Object.freeze({})` module-level.
- `Region.js` `kebab()` regex doesn't handle consecutive caps (`APIKey` → `apikey`). Fine for current vocab, latent.
- `hydrateInline.mjs:85-93` fast-path returns `screen.view._resolved`; slow-path stamps `_resolvedFieldsRef`. Debug consumers see inconsistent shape.
- `tests/runtime/mode-resolution.test.mjs:29-42` use ad-hoc `ok/eq` instead of `node:assert/strict` — loses stack frames on failure.
- `mode-resolution.test.mjs:269-273` reaches into `__test.deepMerge` — couples tests to private impl.
- v3 menu `parent_type=dropdown` still falls back to drilldown (carried from v2 shim). Document or remove.

---

1. **Security trio** — permissions/customizable/null-tombstone. Consumer origins can escalate.
2. **`is_safe_href` protocol-relative.**
3. **`useScreenView` stale-closure + unbounded cache.**
4. **`useMode` hoist to context.** Perf hit per region per URL change.
5. **`BindingsConsumer` rebind-thrash.**
6. **Schema `\$id` reconciliation** across v1/v2/v3.

Test coverage gap: no PHP tests for `Permissions` or null-tombstone merge — both new security-load-bearing.

dabowman:
## Review addendum — WordPress conventions pass

Revalidated against `wordpress-plugin-development` skill (Plugin Handbook + Block Editor Handbook + REST API Handbook). Adds one new finding, refines two recommendations. Original blockers/majors stand.

---

### 🟠 NEW Major — REST `/screen-view` over-discloses to any logged-in user

`includes/class-wp-admin-shell-view-config-rest.php:52-54`:
```php
public static function permission_check() {
    return is_user_logged_in();
}
```

Inline comment claims \"Read-only access is information-only\" — but v3 screens carry `permissions.capabilities` and the resolved view-doc reveals field columns, action labels, filter shapes, hidden screen ids. A subscriber can `GET /wp-admin-shell/v1/screen-view?screen=plugins` and harvest the plugins-screen view-doc without holding `activate_plugins`. Information disclosure + reveals screen ids that would otherwise be cap-gated out of nav.

**Fix.** In `permission_check`, resolve the requested screen's permissions block and gate against `current_user_can()` for the screen's cap floor (route through `WP_Admin_Shell_Permissions::user_can_access_screen()` or equivalent). 403 on failure. Add tests: subscriber requesting admin-only screen-view → 403; admin → 200; unknown screen → 404 not 200.

Pairs with the existing trust-tier blocker — same permission model, different leak surface.

---

### Refinement — replace earlier `trigger_error → error_log` suggestion with `_doing_it_wrong()`

Canonical WP API for developer-error notices. Already escapes internally, respects `WP_DEBUG`, integrated with test harnesses.

Affects:
- `class-wp-admin-shell-menu-items.php:514` (nit in original review)
- `class-wp-admin-shell-v3-compiler.php:294` (slot-route collision Major)

Also a repo-wide pre-existing pattern (4 sites total: `class-wp-admin-shell-{admin-routes,field-collections,menu-items,preload}.php`) — v3 inherits and propagates. Worth a single sweep PR.

---

### Refinement — `is_safe_href` should use WP-native URL validation

Original review suggested adding `if ( strpos( \$href, '//' ) === 0 ) return false;`. Works but rolls own URL validator. WP-native equivalents:

- `wp_validate_redirect( \$href, \$fallback )` — if the href is treated as a redirect target.
- `esc_url_raw( \$href, \$allowed_protocols )` with explicit scheme allowlist via `wp_allowed_protocols()` — for storage.
- For navigation-only (no redirect, no storage): keep current allowlist + add explicit `//` reject, but document the choice.

Pick one path consistently per call site in `class-wp-admin-shell-menu-items.php:480-502`.

---

### Validated as-is against WP non-negotiables

- ✅ `defined( 'ABSPATH' ) || exit;` guard on v3 REST file.
- ✅ `sanitize_callback => 'sanitize_key'` on `screen` param (`view-config-rest.php:39`).
- ✅ Nonce N/A (read-only GET).
- ✅ Hook execution timing — no `init`-only or `plugins_loaded` violations in v3 PHP.
- ❌ `permission_callback` present but **floor too low** — see new Major above.

dabowman:
## Review addendum — Token + CSS pipeline (v3 default shell)

Read-only deep trace of token ingestion → DOM paint for `shells/wp-admin-default-v3.json` + bundled core apps + the three engines.

### Pipeline trace — happy path works

1. **Author seeds** (`shells/wp-admin-default-v3.json:1201-1209`) — three values only: `theme.color.primary`, `theme.color.bg`, `theme.density`. No nested seeds, no slot overrides, no `tokens.*` aliases.
2. **PHP cascade** (`includes/cascade/class-wp-admin-shell-resolver.php:208-233`) — injects engine origin from `src/runtime/engines/core-default/engine.json:210-218`; shell wins overlapping keys.
3. **v3 compiler** (`includes/cascade/class-wp-admin-shell-v3-compiler.php:107-176`) — passes styles through untouched; `WP_Admin_Shell_Tokens::resolve()` loads `core.tokens.json` separately into `window.wpAdminShell.tokens`.
4. **Kernel mount** (`src/runtime/kernel.js:80-85`) — defensive re-merge → `<ThemeProviderHost>`.
5. **Host** (`src/runtime/styles/ThemeProviderHost.js:127-175`) — wraps tree in `data-wpds-theme-provider-id` div, calls `compileStyles(styles, tokens)`.
6. **WpdsThemeProvider** (`src/runtime/styles/WpdsThemeProvider.js:96-120`) — unlocks `@wordpress/theme.ThemeProvider` via `@wordpress/edit-site` allowlist proxy + forwards `{primary, bg}`, `density`. Real provider writes `--wpds-*` CSS vars.
7. **compileStyles** (`src/runtime/engines/core-default/compileStyles.mjs:111-152`) — for v3 shell returns `{top: {}, scoped: [], subtrees: {}}`; `styles.theme` is in `NON_TOKEN_KEYS`, no other slots authored. **`<style>` block emits empty.** Correct for seed-only path.
8. **Region.js** (`src/runtime/regions/Region.js:185, 337-355`) — inline-styles each chrome region with `var(--wp-admin-shell--chrome--…, var(--wpds-…))` chains from `engine.json` template defaults.
9. **Engine CSS** (`src/runtime/engines/core-default/index.css:18-34`) — canvas paints via two-arg fallback. Shell `bg: #1e1e1e` reaches paint through unauthored chrome → WPDS fallback.

**Net:** ThemeProvider does the work; compileStyles is a no-op for v3 default install. Engine CSS interventions are legitimate and minimal.

---

### 🔴 Bugs (kernel DS-neutrality violations)

- **B1. `src/runtime/styles/ThemeProviderHost.js:28` — kernel imports WPDS-specific module.** `WpdsThemeProvider` used as fallback (line 210) + error-boundary recovery. Hypothetical Material engine still drags WPDS into bundle and silently renders under WPDS rules on engine-provider error. Either move `WpdsThemeProvider.js` into `src/runtime/engines/core-default/` + re-export for `core-single-pane` to consume, OR rename → `DefaultThemeProvider.js` + document as kernel's contracted baseline.
- **B2. `src/runtime/kernel.js:7` — `SlotFillProvider` imported from `@wordpress/components`.** Even though Slot/Fill is data infra, per CLAUDE.md the kernel should not pull DS-flavored packages. Non-WPDS engine consumer must load `@wordpress/components` purely to satisfy this provider. Move to engine wrapper or document as platform service via registry hook.
- **B3. `core.tokens.json` is dead weight for v3 default install.** 141-line DTCG tree loaded by PHP, serialized to JS, parsed on every `compileStyles` call — but `shells/wp-admin-default-v3.json` references zero curly-brace aliases (`{color.brand.500}`, `{size.lg}`, etc.). Either (a) author v3 seeds as `{color.brand.500}` aliases to make tokens.json the source of truth, or (b) skip serialization when no aliases present in resolved styles tree + engine declares no token-alias usage.

### 🟠 Unnecessary CSS interventions

- **O1. `src/runtime/engines/core-default/engine.json:140-143, 158-161` — card chrome non-overridable.** `core:main` + `core:detail` paint `border-radius: var(--wpds-border-radius-lg)` + `box-shadow: var(--wpds-elevation-sm)` + `margin` as inline style. No chrome-var escape hatch. Authors can't ship full-bleed layout (single-card / Material consumer) without per-region inline overrides. Wrap each in `var(--wp-admin-shell--chrome--content--card-radius, var(--wpds-border-radius-lg))` etc.
- **O2. `src/runtime/engines/core-desktop/index.css:73, 120` — `rgba(255,255,255,0.18)` and `rgba(255,255,255,0.12)` hardcoded.** Dock separator + tile hover. No var-fallback chain. Chain through chrome slots: `var(--wp-admin-shell--chrome--dock--separator, rgba(255,255,255,0.18))`.
- **O3. `src/runtime/engines/core-desktop/index.css:198-207` — traffic-light colors hardcoded.** macOS conventions defensible, but Linux/Windows/brand-themed consumer has no override. Wrap in chrome vars.
- **O4. `src/runtime/engines/core-default/index.css:115-117` defensive Stack rule** — legitimate per CLAUDE.md (`@wordpress/ui` layered-CSS gotcha). **Keep.**
- **O5. `src/runtime/engines/core-default/index.css:133-155` chrome anchor + svg overrides** — legitimate (cascade-layer fight + `@wordpress/icons` missing `fill=\"currentColor\"`). **Keep.**

### 🟡 Smells

- **S1. `src/runtime/engines/core-desktop/index.css:18, 22, 152-160, 370-388` — hardcoded hex fallbacks (`#0a0a1a`, `#f0f0f5`, `#2a2a3a`, `#404050`, `#1c1c28`) instead of `var(chrome, var(--wpds-…))` chain.** Desktop engine breaks the contract that `core:default` + `core:single-pane` follow consistently. Site overriding WPDS via theme.json sees desktop chrome unchanged. Standardize on the two-arg pattern.
- **S2. `src/apps/desktop-iframe/index.css:17, 37` — `#1c1c28` literal duplicates `core-desktop/index.css:374`.** Two-source maintenance hazard. App-side rule may be entirely redundant — engine paints body bg already at `/body` region scope.
- **S3. `src/apps/navigation/index.css:141-145` `!important` on `ItemGroup` border** — documented in CLAUDE.md. **Only `!important` in entire shell.** Accept.
- **S4. `src/runtime/engines/core-default/engine.json:213-217` ships `theme.color.bg: \"#ffffff\"` as engine default.** Engine description says \"dark chrome\" but engine baseline is white. Confusing — relies on shell to darken. Engine baseline should reflect intended look.
- **S5. Single-seed surface ambiguity.** Shell `bg: #1e1e1e` darkens canvas AND content card (both pull `--wpds-color-bg-surface-neutral` from same WPDS ramp). Author can't have dark canvas + light cards under tier-1 seeds; must drop to tier-3 slot overrides (`styles.color.bg-surface-neutral`). Contradicts spec's \"seeds first\" intent. Either ship a `theme.color.surface` seed alongside `bg`, or document the two-tier-surfaces-require-slots rule.
- **S6. `CHROME_WPDS_BINDINGS` (`compileStyles.mjs:64-89`) has no `canvas` entry.** Only `sidebar`/`toolbar`/`site-hub`. Engine CSS does reference `chrome.canvas.{background,foreground}` correctly via top vars, but `@wordpress/ui` content rendered directly inside `.wp-admin-shell-layout` (future engine modes) wouldn't pick up canvas-scoped WPDS retheming. Add `canvas` entry pre-emptively, or document the omission as intentional.

### ⚪ Nits

- **N1. `ThemeProviderHost.js:150` writes `data-wpds-theme-provider-id` literal in kernel DOM contract.** Leaks WPDS prefix. Rename `data-theme-scope-id` / `data-shell-theme-id`.
- **N2. `core-default/index.css:6` header lies.** Says \"engine-internal layout infrastructure only\"; lines 95-176 ship WPDS-coupling overrides. Both legitimate — fix the header comment.
- **N3. `kernel.js:40, 89` hardcoded `padding: 32` in error fallback div.** DS-neutral but inconsistent with kernel className pattern. Use a className with kernel-CSS-rule structural padding.

---

### Verdict

**Pipeline correct for the seed-only path.** ThemeProvider does the work; compileStyles emits nothing for v3 default. CSS interventions in `core-default` legitimate and minimal — three categories: `@wordpress/ui` layered-CSS workarounds, chrome anchor/svg fixes, engine root paint. Only one `!important` in entire shell, documented.

**Two architectural cleanups outstanding** before claiming kernel-DS-neutrality is complete:
1. Kernel still imports DS-specific modules in two spots (`WpdsThemeProvider`, `SlotFillProvider`) — physically kernel-area, semantically engine.
2. `core:desktop` engine broke from the chrome→WPDS fallback contract that `core:default` + `core:single-pane` follow; ~10 hex literals where the other engines chain through WPDS.

**Two authoring-surface gaps** worth resolving in the v3 spec window:
1. `core:main` / `core:detail` card chrome (`border-radius` / `box-shadow` / `margin`) can't be removed without per-region overrides. No chrome-var escape hatch.
2. Single `bg` seed darkens canvas + cards together — no `surface` seed for two-tier surfaces. Forces tier-3 slot overrides for dark-canvas/light-cards look.

`core.tokens.json` ships but referenced by zero values in v3 default — entire DTCG layer is dead code for the install.

dabowman:
## Review addendum — Phase 3c.2 + delta since prior reviews

Latest pass against `fa67a13`. Skipping items already flagged + still outstanding (see prior addenda).

### ✅ Confirmed fixed since round 1

- **`useScreenView` stale-closure + module-cache structure** (commit `61c41eb`). Renamed to `useDataView`; per-render `cancelled` flag + `cacheKey` re-keying + inflight dedup map. Triple-keyed + screen-keyed lookups now share one cache without collision (`useDataView.js:15-34`).
- **`useMode` recompute thrash** (`12ce8dd` / `7980ca7`). Reads `engineModes` off `useKernel().config` with `window` fallback; `useMemo` over stable refs. Per-region call no longer re-resolves the full mode tree each render.
- **macOS Alt-glyph fallback** (`6389449`). `parseShortcut.mjs:106-120` walks `event.code === 'Key{X}'` when the Option-rewrite hides `event.key`.
- **Cmd+Shift+N browser collision** (`9e6b73a`). `wp-admin-default-v3.json` `navigate-posts-new` shortcut moved to `Mod+Alt+N`. (Note: prior `parseShortcut` Alt-gate concern is now more visible — see below.)
- **DataView 3-axis sanitization.** `(kind, name, variant)` all routed through `WP_Admin_Shell_Data_Field_Collections::sanitize_segment` / `WP_Admin_Shell_Data_View_Config::sanitize_variant_segment` in `class-wp-admin-shell-data-view-rest.php:60-76` + resolver entry points. Addresses prior `fieldsRef` unsanitized major.

### 🟠 New — Phase 3c.2 (`fa67a13`) command-palette rewrite

- **`parseShortcut` Alt-gate still missing** (`parseShortcut.mjs:116-119`). Now load-bearing — `wp-admin-default-v3.json` ships `Mod+Alt+N`, so the `event.code` fallback fires on every `Mod+Alt+<letter>` shortcut. Non-QWERTY layouts where `event.key` for the physical `N` key differs from `n` will spuriously match. One-line fix: `if ( needAlt && /^[a-z]$/.test(keyLower) ) { … }`.

- **`compileCommands.mjs:115` strict `screen.hidden === true`.** Schema enforces boolean, but the cascade runs `wp_admin_shell_data_*` filters on the resolved tree — a plugin callback returning `"hidden": "yes"` / `1` would not hide. Loosen to truthy or document the strict shape as filter-contract.

- **`compileCommands.mjs:118` parameterized-path skip is incomplete.** `path.includes('{') || path.endsWith('/*')` misses mid-pattern wildcards (`/posts/*/foo`) and trailing-slash variants. Latent — no current screens hit these — but the comment claims "Parameterized paths skipped" unconditionally.

- **`compileCommands.mjs:132-135` falls back to `path` when `screen.label` is empty.** Renders palette entry `Go to /posts/new`. Functional but the same code path that produces the screen entry should probably skip the screen when no label is authored, matching the `commands[]` arm's `label === ''` filter at line 60. Symmetry + UX.

- **`useCommandLoader` re-registration churn** (`index.js:89-92`). `commands` array rebuilds via `useMemo` keyed on `[ commandsBlock, screensBlock ]`. Stable from `useKernel()` so OK in practice — but worth noting that any path that flips one of those refs (shell switch, post-mount option-write) will hot-swap the loader. Currently routes-via-reload, so no smoke-test risk; flag if reactive shell-switching lands later.

- **`resolveIcon` silently returns engine fallback for unknown names** (`index.js:60`). The icon registry dev-warns once per unknown name (`iconMap.js`), so the observability already exists — just confirming the palette path doesn't bypass it. No action.

### 🟠 Promotion — prior M2 now actively shipped-broken in v3 default

`shell_allows_user_switch` (`class-wp-admin-shell-resolver.php:307`) reads `$doc['userSwitchable']` (camelCase). `shells/wp-admin-default-v3.json` ships `"user-switchable": true` (kebab, per schema). The default install's user-switching is silently disabled — not just latent. Reclassifying from "latent v2 → v3 inherits" to "blocker for v3 default UX."

### ⚪ Nits

- `command-palette/compileCommands.mjs` test file exists (`tests/runtime/command-palette-compile.test.mjs`) — good. Re-using the `name` field as the identity is implicit; one test asserting `seenNames` collision behavior would lock the contract.
- `useDataView.js:15` cache still unbounded (raised in round 1 — only LRU-able cache left in the v3 surface).

### Verdict delta

Round 1's "security trio" (trust-tier dead code / null-tombstone ungated / customizable pass-through) + `is_safe_href` + REST permission floor + `userSwitchable` typo — **all still outstanding**. Phase 3c.2 didn't touch them.

Net new from 3c.2: one real `parseShortcut` issue now load-bearing (Alt-gate missing → spurious-match risk on shipped `Mod+Alt+N`), three minor `compileCommands` tightening items, one promotion (`userSwitchable` typo now affects shipped v3 default).

dabowman:
## DataViews / DataForm adoption audit

Question: are bundled apps fully using `@wordpress/dataviews` config APIs (both `DataViews` and `DataForm`) where applicable?

**Short answer: DataViews yes, DataForm no.**

### State

Six entity-CRUD apps (`posts`, `comments`, `users`, `plugins`, `taxonomy`, `themes`) consume `<DataViews>` from `@wordpress/dataviews/wp` driven by the C2 dataView config. Good.

**Zero DataForm adopters.** Every edit/create form hand-rolls `InputControl` + `SelectControl` + `useState` + `Stack`. Same package exports `DataForm` — no technical blocker.

### Simplification opportunities

| App | Form | Opportunity |
|-----|------|-------------|
| `taxonomy` | TermEditModal (name/slug/description) | **High** — fields already declared in dataView. Drop modal → `<DataForm fields={fields} onSubmit={save} />` |
| `media` | MediaDetailModal (title/alt/caption/description) | **High** — declare 4-field dataView, swap modal |
| `profile` | ~10 user fields | **Medium** — needs `(root, user)` field collection with edit-shape; could share with users dataView |
| `settings-general` | ~13 site-setting fields + custom date/time radios | **Medium** — synthetic `(root, site)` dataView. Custom radios need field renderer overrides |
| `users` | Bulk-delete-with-reassign | Action modal, not field edit — skip |
| `comments` | Approve/spam/trash | Action callbacks — skip |
| `plugins` / `themes` | None / read-only details | N/A |
| `simple-editor` | Title + blocks | Block editor surface, not DataForm |

### Notable gap

`profile` + `settings-general` skip the dataView registry entirely. Both are entity-edit surfaces (`(root, user)`, `(root, site)`) — natural fit for the 3-axis registry + DataForm.

### Suggested rollout

- **Phase 1 (cheap win, ~100 LOC removed):** taxonomy + media → DataForm. Reuses existing field defs.
- **Phase 2:** register `(root, user)` + `(root, site)` field collections, port profile + settings-general. Needs a field renderer for custom date/time radio compounds.

dabowman:
## DataView config API consumption audit

Question: are we using DataViews + the dataView config API optimally, or leaving maintainability on the table?

**Verdict: not optimal.** All six entity-CRUD apps (`posts`, `users`, `comments`, `plugins`, `taxonomy`, `themes`) re-implement the same plumbing utilities verbatim. Roughly ~650 LOC of non-semantic duplication, plus three config API gaps that force JS workarounds.

### Top 5 refactors (~650 LOC dedup, mechanical)

1. **`compileEligibility()` → `src/runtime/dataView/compileEligibility.js`** — 6 copies, ~15 lines each, identical. Pure utility (turns `eligibleWhen: {status: ['draft','pending']}` into a predicate).
2. **`composeFields(specs, renderers, labels) → DataViews fields[]` → `src/runtime/dataView/composeFields.js`** — 6 copies of identical JSON-spec → DataViews-field translation. App passes renderers map + `FIELD_LABELS`, helper handles assembly (id/type/label/enableSorting/enableHiding/elements/filterBy/render binding).
3. **`composeActions(specs, labels, { callbacks, eligibilityOverrides, iconResolver })`** — 6 copies. Same shape (label/isPrimary/isDestructive/supportsBulk/icon/isEligible/callback|RenderModal).
4. **`useDataViewResync(view, setView, config, dep)`** — 5 copies of the triple-axis-flip reset useEffect. Move into a hook, or fold into `useDataView()` return so consumers stop re-implementing.
5. **`VIEW_DEFAULTS` shape** — extract `VIEW_SHAPE_DEFAULTS` constant; entity-specific sort/perPage migrate fully to `app.json#dataView.defaultView` (mostly there already — apps just stop re-declaring the shape).

### Three JSON-declaration gaps (move from JS → app.json)

1. **`titleField` / `mediaField` strip logic.** Users / posts / taxonomy / themes declare these in JSON, then re-derive in `useMemo` to remove them from visible columns. Should be `useDataView()` return that auto-strips them.
2. **`status` field `elements` fallback.** Four apps hardcode a `STATUS_LABELS → elements` synthesis in JS when JSON omits. Make `elements` required in JSON, drop the fallback.
3. **`defaultLayouts` defaulting.** Every app does `dataViewConfig.defaultLayouts ?? { table: {} }`. Move default into the resolver.

### Three config API gaps (forcing awkward JS)

1. **No JSON syntax for field-presence eligibility.** Plugins hardcodes `(item) => !!item.pluginUri`. Extend schema: `eligibleWhen: { field: "pluginUri", operator: "exists" }`.
2. **`RenderModal` callback can't be JSON-declared.** Every trash/delete/edit modal lives in a JS branch like `if (spec.id === 'trash')`. Add a keyed modal-template registry or convention — app exports `modals[id]` map, helper wires by id.
3. **REST-vs-client filtering not declarable.** Plugins' `/wp/v2/plugins` doesn't honor `per_page/page/search/status` — app does client-side filter. dataView config should declare `query.transport: "rest" | "client"` so apps stop re-implementing pagination math per endpoint.

### Duplication summary

| Issue | Scope | LOC saved |
|-------|-------|-----------|
| `compileEligibility()` duplication | 6 apps × ~15 lines | ~90 |
| `buildFields()` duplication | 6 apps × ~30 lines | ~180 |
| `buildActions()` duplication | 6 apps × ~40 lines | ~240 |
| `VIEW_DEFAULTS` duplication | 6 apps × ~8 lines | ~48 |
| useEffect resync duplication | 5 apps × ~10 lines | ~50 |
| `titleField` strip logic duplication | 3 apps × ~15 lines | ~45 |
| **Total extractable** | | **~650** |

### Bottom line

`useDataView()` resolves the doc cleanly — the cascade + 3-axis lookup work. Apps **consume** it sub-optimally: every app re-implements the same 5 utilities, and three patterns that should be JSON-declared live as branchy JS instead.

Suggested rollout:
- **Phase 1 (mechanical, no schema change):** extract the 5 utilities into `src/runtime/dataView/*.mjs`. ~600 LOC removed; lockstep pattern across all six apps.
- **Phase 2 (schema additive):** add `eligibleWhen` operator vocabulary, modal-template registry, `query.transport` declaration. Kills remaining JS hacks.

PR #52, opened 2026-05-20
v3: token + CSS pipeline cleanup — kernel DS-neutrality + dead-token short-circuit

dabowman:
Addresses the **Token + CSS pipeline review** on PR #49 (comment [4500723352](https://github.com/dabowman/WordPress-Admin-Environment/pull/49#issuecomment-4500723352)). Scope locked to 🔴 bugs + 🟠 major + 🟡 smells; nits and accepted items deferred.

## Kernel DS-neutrality

- **B1**: `WpdsThemeProvider` moved out of `src/runtime/styles/` into `src/runtime/engines/core-default/`. `core-single-pane` and `core-desktop` import it from there (shared bundled-engine helper). Kernel imports no DS-specific module. `ThemeProviderHost` no longer falls back to WPDS on engine-provider crash — renders children pass-through and logs. A hypothetical non-WPDS engine plugin is unaffected by anything in the bundled engines.
- **B2**: `SlotFillProvider` moved from `kernel.js` into each bundled engine's `Layout` component. `@wordpress/components` is a DS-flavored package; a non-WPDS engine that doesn't ship `Slot`-consuming apps can omit the wrap entirely.
- `core-desktop`'s `EngineSource` now declares `ThemeProvider: WpdsThemeProvider` explicitly (was relying on the kernel fallback that has been removed).

## Token pipeline

- **B3**: PHP `wp_admin_shell_styles_reference_tokens()` detects whether the resolved styles tree references any foreign (non-`{styles.*}`) alias. When none, token serialization is skipped — the DTCG layer is no longer dead weight on the wire for seed+slot-only installs. v3 default ships zero foreign aliases, so the tokens payload now drops from the page entirely.

## CSS interventions

- **O1**: `core:main` and `core:detail` expose `--wp-admin-shell--chrome--content--card-{radius,shadow,margin}` slots alongside the existing `card-background` slot. Authors override card chrome without per-region inline rules.
- **O2 / O3 / S1**: `core-desktop/index.css` — hardcoded rgba/hex fallbacks now chain through WPDS surface tokens (the `var(chrome, var(--wpds-…))` pattern `core:default` + `core:single-pane` follow). Dock separator + tile hover, traffic-light controls, window-frame background/foreground/border, window body chrome. Engine identity preserved via `engine.json#default-styles.chrome.*` cascade.
- **S2**: `desktop-iframe` app drops duplicate `#1c1c28` fallbacks. Window body region paints the surface; the iframe wrapper is transparent.
- **S4 / S5**: `core:default` engine baseline ships "dark chrome + light cards" identity (`chrome.{canvas,sidebar,toolbar}.background=#1e1e1e` with foreground `#e0e0e0`; `theme.color.bg` stays `#ffffff` so the WPDS surface ramp keeps cards bright). v3 default shell drops its now-redundant `bg:#1e1e1e` override.
- **S6**: `canvas` added to `CHROME_WPDS_BINDINGS` — `@wordpress/ui` content rendered directly under `.wp-admin-shell-layout` rethemes automatically when chrome canvas slots are authored.

## Other

- **N1**: `data-wpds-theme-provider-id` renamed to `data-shell-theme-id` (and `data-wpds-shell-detail` to `data-shell-theme-detail`) so the kernel DOM contract carries no WPDS prefix. Tightly coupled with B1 — keeping the WPDS name would contradict the kernel-DS-neutrality story.
- `CLAUDE.md` theming section updated. `docs/engines-and-design-systems` + `docs/v1-token-emission` attribute-name refresh.

## Test plan

- [x] `npm run test:runtime` — 27/27 pass
- [x] `npm run test:schema` — 101/101 pass
- [x] `npm run test:parity` — 4/4 pass
- [x] `npm run test:engines` — 77/77 pass
- [x] `npm run build` — succeeds, no asset budget regressions beyond pre-existing 1.75 MiB vendors chunk
- [x] Token detector validated standalone (10 cases — empty / no-alias / hex / within-doc / foreign / deep / mixed)
- [ ] PHP tests via `wp-env` (wp-env not available in this environment; PHP `-l` syntax check passes)
- [ ] Manual smoke: each of the five bundled shells renders + cap-gating per role + token alias resolution end-to-end

## Items deferred (not in this PR)

- **N2 / N3** — comment-header lie in `core-default/index.css` + hardcoded `padding: 32` in `kernel.js` error fallbacks. Nits per the review scope vote.
- **O4 / O5** — the defensive `[class*="__stack"]` rule + chrome anchor/svg overrides. Accepted as legitimate per the review.
- **S3** — single `!important` in the shell (`ItemGroup` border in `navigation/index.css`). Accepted.

https://claude.ai/code/session_01JPptLjQhXWeHW5zy1kyJbb

---
_Generated by [Claude Code](https://claude.ai/code/session_01JPptLjQhXWeHW5zy1kyJbb)_


dabowman:
## Code Review — Token + CSS pipeline cleanup

Verified diff against the original review (#49 comment 4500723352). Kernel DS-neutrality cleanup (B1, B2) lands clean and `grep` confirms no `@wordpress/components` / `@wordpress/ui` / `--wpds-*` / chrome-class references survive in `src/runtime/*` outside `engines/`. B3 short-circuit, O1 card-chrome slots, O2/O3/S1 desktop-CSS chain, S6 canvas binding, N1 attribute rename — all present and structurally correct. Lint + 4 Node test suites pass locally. **However**, the S4/S5/S6 combination introduces a real regression in the very identity it claims to establish.

---

### 🔴 Bugs

- **S6 + S4 combine to break "dark chrome + light cards"** identity the PR claims to ship. `core-default/compileStyles.mjs:64-71` now emits `--wpds-color-bg-surface-neutral: #1e1e1e` scoped to `.wp-admin-shell-layout` whenever `chrome.canvas.background` is set. Since `engine.json#default-styles.chrome.canvas.background = #1e1e1e` (line 220), the scoped override fires for every install of the engine. `core:main` card paints `background: var(--wp-admin-shell--chrome--content--card-background, var(--wpds-color-bg-surface-neutral))` (engine.json:140) — chrome var is unauthored → falls through → resolves to `#1e1e1e` under the scope → **content cards now paint dark**, not white. The same trap hits `core:detail` (line 158) and any other descendant of `.wp-admin-shell-layout` reading the WPDS surface ramp.

  Reproduction (Node):

  ```
  compileStyles({theme:{color:{bg:'#fff'}}, chrome:{canvas:{background:'#1e1e1e',foreground:'#e0e0e0'}}}, {})
  → scoped[0] = { selector: '.wp-admin-shell-layout', vars: { '--wpds-color-bg-surface-neutral': '#1e1e1e' } }
  ```

  Fix options:
  - Drop `background` from the `canvas` binding in `CHROME_WPDS_BINDINGS`. Authors who want `@wordpress/ui` content rendered directly under `.wp-admin-shell-layout` to retheme can do so via a dedicated `--wpds-color-bg-canvas` token, not by hijacking the `surface-neutral` ramp the cards consume.
  - Or scope the canvas WPDS override to `.wp-admin-shell-layout > :is(...)` to avoid descendant inheritance into card content. Coarse — fragile to layout changes.
  - Or pin card backgrounds to a separate chrome variable in `engine.json#default-styles.chrome.content.card-background` so the cascade stops relying on the WPDS-ramp fallback path.

  Recommend option 3 + drop `background` from canvas binding — restores the previously-working identity and makes the "dark chrome + light cards" claim load-bearing instead of accidental.

  Add a runtime test: `tests/runtime/compile-styles-tokens.test.mjs` should cover the canvas-binding + content-card interaction so this can't silently regress again.

- **S5 not actually fixed.** The PR description claims `theme.color.bg=#ffffff` "keeps the WPDS surface ramp light so cards paint bright." That's only true at the WPDS-provider scope; the new canvas binding then re-darkens the surface token at `.wp-admin-shell-layout` scope, which is the parent of every card. Functionally S5 is still open — the seed-vs-slot ambiguity the original review flagged remains, just routed through a different vector.

### 🟠 Major

- **`tokens` payload shape inconsistency.** `wp-admin-shell.php:410-412` emits `array()` when no foreign aliases are present. `wp_json_encode( array() )` produces `[]`, not `{}`. JS consumers (`compileStyles.mjs:161-171`) tolerate it via `typeof [] === 'object'`, but the kernel + every downstream consumer typedef the field as `Object`. Either return `(object) array()` (serializes to `{}`) or normalize on the JS side. Latent shape-drift hazard for any future code that does `Object.keys(tokens)` early.

- **B3 detector scope is `styles` only.** `wp_admin_shell_tree_has_token_alias()` walks `$config['styles']` exclusively. App manifests `app.json#dataView` documents, region template `default-style` values, and any per-app `styles` block (none today; the schema doesn't forbid future use) won't be considered. Tighten the contract — either document "DTCG aliases only allowed under `styles.*`" in the schema + reject on validation, or extend the scan. Today this is safe because the v3 default ships zero alias usage anywhere; it's a documentation/contract gap, not a live bug.

- **Three engines now have a duplicate `WpdsThemeProvider` import path through `core-default/`.** B1's resolution is structurally correct (DS-specific module lives with the engine that contributes it) but `core-single-pane` and `core-desktop` reaching across to `core-default/WpdsThemeProvider` couples three engines tightly. If a non-WPDS engine plugin ships and a user activates it alongside `core-default`, the bundled WPDS provider is still loaded by the other two bundled engines. Acceptable for now (all three are first-party WPDS-flavored), but consider promoting `WpdsThemeProvider` into a separate shared "wpds-bundled-engines-shared/" subdirectory so the contract — "any engine importing this opts into WPDS" — is explicit rather than incidental. Not blocking.

### 🟡 Smells

- **`desktop-iframe` loading state now relies on transparent iframe default.** S2 dropped the wrapper background. Modern browsers default `<iframe>` to transparent, so the parent's window body paint shows through during load — but this is a behavioral contract change worth a manual smoke pass before tagging. If any environment renders iframes with a UA-default white background, the loading-state overlay (`position: absolute; inset: 0`) sits on top of white instead of the chrome surface. Lower priority — works in Chrome/Safari/Firefox per current behavior.

- **`docs/wordpress-design-tokens-catalog.md:162` still references `data-wpds-theme-provider-id`.** N1's rename swept four files but missed this one. The reference there is to WPDS's **internal** wrapper attribute (which the real `@wordpress/theme.ThemeProvider` emits and which the shell doesn't control), so it's accurate-for-WPDS-but-confusing — worth a footnote disambiguating the two attributes (shell-host outer wrapper = `data-shell-theme-id`; WPDS inner wrapper = `data-wpds-theme-provider-id`).

- **`engine.json#default-styles.chrome.site-hub` is foreground-only** (line 232). Sidebar / toolbar / canvas all ship background + foreground; site-hub ships only foreground. Probably intentional (site-hub inherits from sidebar via DOM nesting) but worth an inline comment so future contributors don't "fix" it.

### ⚪ Nits

- Removed `wpdsThemeProviderAvailable` named export from `WpdsThemeProvider.js` — grep confirms no consumers. Good.
- Defer items (N2, N3, O4, O5, S3) all correctly out of scope per the PR description.
- Original review items missed by this PR are out of scope (the review explicitly locked to "Token + CSS pipeline review"). However, the "security trio" (trust-tier, null-tombstone, customizable pass-through) from PR #49 round 1 remains outstanding on the v3 branch and is a higher merge-gate concern than anything in this PR.

---

### Verdict

**Hold for one targeted fix.** Lint + Node tests pass and B1/B2/N1 are clean wins. The S6 canvas-binding addition + S4 chrome.canvas.background engine default combine to break the identity the PR claims to establish; cards inherit `#1e1e1e` instead of `#ffffff`. Fix that one interaction (drop `background` from canvas binding OR author a `chrome.content.card-background` slot in engine defaults) + add a compile-styles regression test, and the PR is good to merge.

### Top fix priorities

1. **S6 canvas-binding cascade interaction** — content cards now paint dark. Single biggest item.
2. **`tokens` shape consistency** — `array()` → `(object) array()` in `wp-admin-shell.php:411`.
3. **B3 scope documentation / extension** — clarify the "aliases only under `styles.*`" contract.
4. **Manual smoke** — five bundled shells × three engines × cap-gated roles. The PR description marks this checkbox unchecked; the regression above will surface immediately on `wp-admin-default-v3` mount.

---

dabowman:
Superseded by #67. Validated all 22 files against current `main`: the structural half of this PR (B1 WpdsThemeProvider relocation, B2 SlotFillProvider-into-engine) already shipped via the DS-decoupling refactor, and N1's attr rename was superseded (main uses the cleaner `data-theme-scope-id`, not `data-shell-theme-id`). The remaining token/CSS hygiene (B3, O1, S4/S5, S6, O2/O3/S1/S2) is salvaged and rebased onto `main` in #67. Closing — base `feat/wp-admin-shell-v3` is stale + conflicting.

PR #54, opened 2026-05-20
v3: 3c.3 — classic wp-admin menu bridge

dabowman:
## Summary

- Phase 3c.3 of `docs/v3/roadmap.md`. New `WP_Admin_Shell_Classic_Menu_Bridge` PHP class walks `\$GLOBALS['menu']` + `\$GLOBALS['submenu']` and synthesizes v3 admin.json entries (`screens[ingested-<slug>]` + `menu.ingested.items[<id>]`) for third-party plugin menu registrations.
- Core wp-admin entries skipped via a static slug list (`index.php`, `edit.php`, `upload.php`, etc.) — the shell already provides v3 screens for them.
- `wp_admin_shell_classic_menu_core_slugs` filter expands the skip list for sites/plugins that want the shell to cover additional slugs natively.
- Mirrors the 3c.1 dashboard-widgets pattern: cascade contribution at `wp_admin_shell_data_plugin` priority 6, idempotency guard against double-apply, derive_entry_id namespace-safe.
- Bridge contributes a `classic_menu_bridge` signal to `wp_admin_shell_cache_signals` so plugin (de)activation between requests picks a different cache bucket.

## What changed

| Layer | Files |
|---|---|
| PHP bridge | `includes/cascade/class-wp-admin-shell-classic-menu-bridge.php` (new, 480 lines) — `scan()`, `is_core_slug()`, `derive_path()`, `derive_screen_id()`, `map_icon()`. Cascade hook at file load. |
| Bootstrap | `wp-admin-shell.php` — `require_once` added after menu-items + admin-routes shims |
| Tests | `tests/php/run-classic-menu-bridge-tests.php` (new, **63 assertions**) — covers core-skip, third-party ingest, submenu nesting under plugin parent + under core parent, icon mapping (dashicons + data-URI + empty fallback), capability propagation, slug→path derivation, id collision with v3 screens, admin.json override precedence, filter-based core-slug expansion, idempotency under double-apply, empty `\$GLOBALS['menu']`, container creation + preservation, label `<span>` stripping. |
| Docs | `docs/v3/roadmap.md` 3c.3 marked ✅ Shipped; `docs/wp-admin-shell-design-spec.md` §13 entry #13 added (classic menu bridge); old #13 → #14 |

## Design decisions

1. **Static core-slug list + filter expansion.** Survey alternative was reflection over registered callback file paths — slower + brittle. Static list with `wp_admin_shell_classic_menu_core_slugs` filter is simpler + extensible.
2. **Eager scan at filter time** (not lazy). The filter runs only when the cascade resolver builds a config for the shell page — already the right gating point. No need to defer to first-mount.
3. **Default container hardcoded `menu.ingested.items`** with "Plugins" label fallback. Admin.json can override the container label at any cascade origin; the bridge only writes to `items`, never clobbers `label`.
4. **App id = `iframe:<original-slug>`** verbatim. Schema's `namespacedId` regex accepts query-stringed slugs (`iframe:edit.php?post_type=product`).
5. **Synthesized container screens are `hidden: true`** — they have no `path`/`app`, just a menu placement; `hidden: true` prevents resolver orphan-screen warnings.
6. **Label `<span class='update-plugins count-N'>...</span>` stripped** — wp-admin emits update-count bubbles inline; bridge cleans them before storing.
7. **`reset()` is a no-op** (documented). Bridge holds no static state; method kept for API symmetry with other registries so test harnesses can reset uniformly.

## Test plan

- [x] `npm run test:schema` PASS 101 / FAIL 0
- [x] `npm run test:runtime` PASS (full suite, last block 27/27)
- [x] `npm run test:engines` PASS 77
- [x] `npm run test:parity` PASS 4
- [x] `npm run lint:js` / `lint:ts` clean
- [x] `npm run build` clean
- [x] PHP `run-classic-menu-bridge-tests.php`: PASS 63
- [x] All regression PHP suites green (menu-route-shims 70, dashboard-widgets 56, cascade 39, shape 126, data-view 98, v3-compiler 53, etc.)
- [ ] **Browser smoke deferred** — purely server-side feature; output flows through tested cascade pipeline (`screens[]` + `menu.<id>.items` shapes the runtime React tree already consumes). The `iframe:<slug>` app id is what `wp-admin-default-v3.json` already ships for ~a dozen legacy wp-admin pages.

## After merge

Phase 3c queue close to wrap:
- **3c.4** Multi-app layout algorithm (~4-6 hours per survey — smaller than roadmap's 1-2 day estimate; compiler-change-only + slot templates already exist).
- Then **3d.1** migrate 5 remaining bundled shells, **3d.2** v2→v3 migration helper, **3d.3** test surface rewrites, **3d.4** doc sweep.


dabowman:
## Code review

Phase 3c.3. Cascade pattern, idempotency, cache signal, test coverage all match the 3c.1 / C3 / C4 precedent. 63 PHP assertions cover the right surfaces.

### Issues

**1. `iframe:<slug>` app id not resolvable at mount time (latent across v3, not new to this PR).**

`src/runtime/regions/mountApp.js:342` `resolveAppInstance` accepts only `core:*` / `plugin:*` namespaces — `iframe:<slug>` returns `null`. No PHP compiler translates `iframe:<slug>` → `core:iframe-fallback` + `config.url` either. `IframeApp` reads `config.url` exclusively; without it, `rawUrl === ''` early-returns `null`. Existing `wp-admin-default-v3.json` uses the same `iframe:update-core.php` shape — so the bug is latent across v3, but this PR multiplies it across every ingested third-party plugin entry. "Browser smoke deferred" in the test plan masks it.

Fix options:
- Add `iframe:` prefix branch in `resolveAppInstance` returning `{ id: appRef, source: 'core:iframe-fallback', config: { url: appRef.slice(7) } }`.
- Or transform at `WP_Admin_Shell_V3_Compiler::synthesize_routes()` time — detect `iframe:` prefix, rewrite to `core:iframe-fallback` + merged `config.url`.

Recommend compiler-side rewrite. Keeps mount path DS-neutral, fixes existing v3 default shell with no kernel change.

**2. `is_core_slug()` re-runs `apply_filters('wp_admin_shell_classic_menu_core_slugs', …)` on every call.**

Once per `$menu` entry + once per submenu child = 30+ filter invocations per `scan()` on a typical site. Memoize the filtered list via request-scoped static cache.

**3. Cache-signal callback runs full `scan()` on every cascade cache-key computation.**

Bridge already calls `scan()` inside `contribute()`. Memoize the result on a request-scoped static so cache-signal + contribute share one walk.

**4. Dead `$parent_slug` parameter in `scan_children`.**

`unset( $parent_slug ); // Reserved for future use…` — YAGNI. Drop until a consumer needs it. CLAUDE.md convention: no scaffolding for hypothetical futures.

**5. Filter result not shape-validated.**

`apply_filters('wp_admin_shell_classic_menu_core_slugs', self::$CORE_SLUGS)` falls back if return isn't an array, but a callback returning `['edit.php', 42, false]` slips through `in_array(..., true)` silently. `array_filter($core_slugs, 'is_string')` catches misuse early.

**6. Hidden synthesized container screen — `icon` written but unread.**

`screens['ingested-tools-php']` carries `icon` while `hidden: true`. Menu item also carries `icon`. Admin.json override on the menu item leaves the screen icon stale. Skip writing `icon` onto the hidden container screen.

**7. Coexistence test doesn't probe the real collision case.**

Test asserts manual item at `menu.manual-item` (top-level) + bridge item at `menu.ingested.items.ingested-my-plugin-page` — different containers, trivially passes. Real case: manual `register_menu_item` writes into `menu.ingested.items.ingested-my-plugin-page` and the bridge skips. Add that assertion.

**8. Asymmetric empty-slug handling.**

`derive_path('')` → `/admin`. `derive_screen_id('')` → `ingested-unknown`. Both publicly callable. Either both reject or both return sentinels. `contribute()` guards upstream so defensive only, but the public API surface is inconsistent.

### Nits

- `'icon' => $icon !== null ? $icon : 'menu'` → `self::map_icon($wp_icon) ?? 'menu'`.
- Spec §13 renumbered #13 → #14; CLAUDE.md still says "Thirteen extension surfaces" / lists #13 as dashboard widget. Update CLAUDE.md in this PR or a follow-up.
- `reset()` no-op kept "for API symmetry" — fine, documented.

### Risk summary

- **Highest:** issue #1 — bridge will produce non-rendering screens in browser. Block merge until the `iframe:<slug>` resolver / compiler transform lands.
- Medium: #2 / #3 filter + scan re-evaluation.
- Low: rest.

### Recommendation

Hold merge pending the `iframe:` resolver. Other items follow-up-friendly.

dabowman:
Review addressed in \`dd45435\`. All 8 issues + 2 nits + iframe translator tests landed.

### Blocker — #1 \`iframe:<slug>\` resolver

Fixed at v3 compiler. New \`translate_iframe_app_refs()\` pass runs at the end of \`compile()\` on both v2 + v3 paths. Walks screens (shorthand \`app\` + long-form \`apps[]\`) and the synthesized \`routes\` block; rewrites \`iframe:<slug>\` → \`core:iframe-fallback\` with \`config.url: <slug>\`. Idempotent. Author-supplied \`config.url\` survives (explicit overrides win).

Authors keep emitting \`iframe:<slug>\` shorthand in shells/manifests/PHP shims; the compiler unifies onto the single \`core:iframe-fallback\` mount path. **Fixes the latent v3-wide bug** (\`wp-admin-default-v3.json\` was shipping broken \`iframe:update-core.php\` etc.), not just the bridge surface.

10 new assertions in \`run-v3-compiler-tests.php\` cover: screens shorthand, long-form \`apps[]\` with slot preserved, \`core:posts\` siblings unchanged, author \`config.url\` override survival, routes block, idempotency, v2 path (\`synthesize_v2_screens_from_routes\` → rewritten).

### Follow-ups

- **#2 — \`is_core_slug()\` memoized.** New class-static \`$core_slugs_cache\` stores the filtered list per-request. \`reset()\` drains for tests.
- **#3 — \`scan()\` memoized.** Static cache keyed on \`md5(serialize($menu, $submenu))\`; second call within a request returns cached records.
- **#4 — Dead \`$parent_slug\` param dropped.**
- **#5 — Filter result shape-validated** via \`array_filter($filtered, 'is_string')\`. Non-string entries can't poison \`in_array\`.
- **#6 — Hidden synthesized container screens no longer write \`icon\`.** Authors override icon at \`menu.ingested.items[<id>]\` (single source of truth — no menu/screen drift).
- **#7 — Real id-collision test added.** Pre-seeds \`menu.ingested.items.ingested-my-plugin-page\` and asserts first-write wins (custom label survives bridge pass).
- **#8 — Empty-slug symmetric.** Both \`derive_path('')\` and \`derive_screen_id('')\` return empty string now (was mixed \`/admin\` + \`ingested-unknown\`).

### Nits

- \`?? 'menu'\` instead of ternary.
- \`CLAUDE.md\` "Thirteen" → "Fourteen" + new bullet #14 documenting the bridge's \`wp_admin_shell_classic_menu_core_slugs\` filter + \`iframe:<slug>\` emission contract.

### Test deltas (all green)

- PHP \`run-classic-menu-bridge-tests.php\`: 63 → **66** (+3: symmetric empty-slug × 2, real-collision × 1).
- PHP \`run-v3-compiler-tests.php\`: 53 → **65** (+12: iframe translator coverage).
- All other PHP suites unchanged + green.
- \`npm run test:schema\`: 101.
- Lint + build clean.

PR ready for re-review.

dabowman:
## Re-review — `dd45435`

All 8 items addressed. Verified each against the diff.

### Verified

- **#1 `iframe:<slug>` translator.** New `WP_Admin_Shell_V3_Compiler::translate_iframe_app_refs()` runs at end of `compile()` on both v2 + v3 paths. Walks screens shorthand `app`, long-form `apps[]`, and synthesized `routes`. Rewrites prefix → `core:iframe-fallback` + `config.url: <slug>`. Author-supplied `config.url` preserved (only fills empty slot). Idempotent (second pass no-ops because `iframe:` prefix gone). 12 new compiler assertions cover all paths including the v2-path screen synthesis fork. Right place — keeps mount path DS-neutral, fixes existing v3 default shell at the same time.

- **#2 `is_core_slug()` memoization.** `core_slugs_cache` class-static drains via `reset()`. Filter dispatch drops from 30+/scan to 1/request.

- **#3 `scan()` memoization.** Function-static signature-keyed on `md5(serialize([$menu, $submenu]))`. Cache-signal + contribute share one walk per request. Signature recomputes on globals mutation so test harnesses see fresh scans after `reset_globals`.

- **#4** `$parent_slug` param dropped from `scan_children`.

- **#5** `array_filter($filtered, 'is_string')` + `array_values()` shape-validates filter result.

- **#6** Hidden synthesized container screen no longer carries `icon` — authors override at `menu.ingested.items[<id>]`.

- **#7** Real collision test added: pre-seeds `menu.ingested.items.ingested-my-plugin-page` w/ custom label, asserts first-write wins through `apply_filters('wp_admin_shell_data_plugin', ...)`.

- **#8** `derive_path('')` + `derive_screen_id('')` both return `''`. Symmetric.

### Minor observations (non-blocking)

- **`reset()` save/empty/scan/restore globals dance** to drain the function-static memo is awkward. Function-statics inside `scan()` can't be externally cleared — only by re-invoking `scan()` with empty input to flip the signature. Cleaner to hoist the two memo slots to class-statics (matches `core_slugs_cache` pattern), but the current workaround works + is documented.

- **`md5(serialize(...))` for cache signature** — fine on flat string arrays. If wp-admin ever stuffs callbacks/closures into `$GLOBALS['menu']` (it doesn't), serialize would throw. Defensive surface, low risk.

- **Edge case untested:** `app: "iframe:"` with empty slug → translator sets `config.url = ''` → IframeApp early-returns `null`. Defensive only; no current emitter produces this. Could add a one-line `if ($slug === '') continue;` in the rewriter to skip cleanly. Optional.

- **`derive_path('')` → `''` is technically schema-invalid** (path pattern requires `^/`). Only reachable defensively (contribute() guards `$slug === ''` upstream). Documented in comment. Fine.

### Verdict

LGTM. Ready to merge.

PR #55, opened 2026-05-21
v3: 3c.4 — multi-app layout (slot-keyed route synthesis)

dabowman:
## Summary

- Phase 3c.4 of `docs/v3/roadmap.md`. Multi-app screens (e.g. `posts` + paired `core:editor` in a `slot: "detail"` peer) now wire end-to-end.
- Compiler's `synthesize_routes()` walks every `apps[]` entry. Non-primary entries with a `slot` emit `@<slot>/<path>` slot-namespaced routes; engine regions declaring `routing.route-key: "<slot>"` pick them up via the existing `useRouteForRegion` infrastructure.
- `core:default` engine `defaultRegions` gains a `detail` region using the pre-existing `core:detail` template.
- Kernel `<PersistentRegion>` emits `data-app-mounted="true"|"false"` (DS-neutral data attribute); engine CSS collapses empty `detail` regions on single-app screens.
- Estimate: roadmap said 1-2d; landed in ~4-6h per the survey's lower-bound prediction. Routing infrastructure already worked end-to-end — only the compiler's primary-only walk was the missing piece.

## What changed

| Layer | Files |
|---|---|
| Compiler | `includes/cascade/class-wp-admin-shell-v3-compiler.php` — `synthesize_routes()` extended to walk all `apps[]` entries. Per-entry `slot` produces `@<slot>/<path>` route; `_self` on non-primary entries guards against primary-clobber. Existing routes win on collision. |
| Engine | `src/runtime/engines/core-default/engine.json` — `detail` added to `defaultRegions` with `routing.route-key: "detail"`. |
| Kernel | `src/runtime/regions/Region.js` `<PersistentRegion>` emits `data-app-mounted` data attr. DS-neutral. |
| Engine CSS | `src/runtime/engines/core-default/index.css` — collapses empty detail region via the data attr. |
| Tests | `tests/php/run-v3-compiler-tests.php` +14 assertions covering primary survival, detail slot synthesis, screenId injection, per-entry config preservation, triple-app screens, no-slot peers, single-app unaffected, escape-hatch routes win on collision. |
| Docs | Roadmap §3c.4 marked ✅ Shipped with design notes. |

## Slot-routing model

```jsonc
// Two-app split screen
"split": {
  "label": "Split",
  "path": "/split",
  "apps": [
    { "id": "main", "app": "core:posts" },             // mounts at /split (route-key: _self)
    { "id": "peer", "app": "core:editor", "slot": "detail" }  // mounts at @detail/split
  ]
}
```

Compiler emits:

```jsonc
"routes": {
  "/split": { "app": "core:posts", "config": { "screenId": "split" } },
  "@detail/split": { "app": "core:editor", "config": { "screenId": "split" } }
}
```

Engine's `detail` region reads `@detail/<current-primary-path>` and mounts `core:editor`. Single-app screens (no `slot: "detail"` entry) leave the detail route slot empty; `<PersistentRegion>` emits `data-app-mounted="false"`; CSS collapses the container.

## Test plan

- [x] `npm run test:schema` PASS 101 / FAIL 0
- [x] `npm run test:runtime` all green
- [x] `npm run test:engines` PASS 77
- [x] `npm run test:parity` PASS 4
- [x] `npm run lint:js` / `lint:ts` clean
- [x] `npm run build` clean
- [x] PHP `run-v3-compiler-tests.php`: PASS 79 (was 65, +14)
- [x] PHP `run-shape-tests.php`: PASS 128 (was 126, +2 — new detail region surfaces in resolved-shape sweeps)
- [x] PHP `run-engine-defaults-tests.php`: PASS 22 (unchanged)
- [ ] **Browser smoke #1 — single-app screen.** Activate `wp-admin-default-v3`. Navigate to `/posts` (single-app). Detail region absent; content takes full width.
- [ ] **Browser smoke #2 — multi-app screen.** Author a v3 shell with a split-pane screen (`apps: [{id:'main',app:'core:posts'}, {id:'peer',app:'core:editor',slot:'detail'}]`). Both mount side-by-side; detail panel takes ~50% width.

## After merge

Phase 3c **complete**. Queue moves to 3d:
- **3d.1** Migrate 5 remaining bundled shells (drop v2 shells).
- **3d.2** v2 → v3 migration helper CLI.
- **3d.3** Test surface rewrites.
- **3d.4** Documentation sweep.
- **3d.5** Pre-v3.1 shim-removal hardening.


dabowman (commented):
## 🚨 Blocker — synthesized routes unreachable from JS runtime

Compiler emits `@detail/split` route keys. JS matcher cannot consume:

- `matchRoute.mjs:27` — `PATTERN_VALID_RE = /^\/[A-Za-z0-9_/{}\-*]*$/`. `@` not in char class, pattern must start with `/`. `isValidRoutePattern("@detail/split")` → false → `matchPattern` returns null every call.
- `router.js:129 useRouteForRegion` → `readSlot(url, "detail")` (matchRoute.mjs:214) reads URL **query param** `?detail=...`. For `#/split` (browser-smoke #2 case), slot value is `""` → no match → region empty.
- No code populates `?detail=/split` when primary path is `/split`. No code strips `@<slot>/` prefix before matching. No code rewrites slot value to `@<slot>/<primary>` before lookup.

Net: detail region renders empty for every URL on every screen PR claims wires up. PHP unit tests pass because they assert compiler output dict only; consumption side unverified. Browser smoke #2 unchecked — that's the gap.

Three possible fixes (each needs new runtime tests):

1. Widen `PATTERN_VALID_RE` to accept `@<slot>/...`; have `useRouteForRegion` synthesize lookup value as `@<route-key>/<primary>` from primary path + region's route-key, not URL query param.
2. Router-side rewrite: auto-populate `?detail=<primary>` when `@detail/<primary>` route exists for current primary path.
3. Strip `@<slot>/` prefix in `matchRoute` when region's route-key matches `<slot>`, match remainder against slot value.

Option 1 most consistent with spec §6.4 framing of regions as URL-slot consumers; avoids polluting URL with redundant query params.

## Secondary issues

- **`primary_path` derivation duplicates existing logic.** `synthesize_routes` computed equivalent value at line 670 (`$path !== '' ? ltrim( $path, '/' ) : $screen_id`). New code at line 686 reintroduces same expression as `$primary_path`. Reuse `$path` + match the single-line form used above for slot route synthesis.
- **Apps-list intra-collision untested.** Two non-primary entries sharing one slot silently first-wins via `! isset` guard. Either test first-wins is the contract, or log + warn on duplicate slot. Plugin authors who collide see one entry vanish with no signal.
- **`_self` defensive guard unverified.** Comment claims defensive intent at line 706, no assertion proves non-primary entry with `slot: "_self"` skipped instead of clobbering primary. Add one assertion.
- **Path with parameter segments (e.g. `/posts/{id}/edit`) untested.** Slot key becomes `@detail/posts/{id}/edit` — interaction with `interpolate(config, params)` after route match unclear. Add screen with parametric path + `slot: "detail"` peer.
- **CSS rule per-region-id hardcoded.** `[data-region-id="detail"][data-app-mounted="false"]` covers `detail` only; future `inspector` / `preview` need parallel rules. Generic `[data-app-mounted="false"]` on slot-routable regions (or class added by engine template) scales.
- **Roadmap header reads `(~1-2 days, may be deferred) — ✅ Shipped`.** Drop "may be deferred" parenthetical post-ship — line reads stale on later browse.

## Style + convention

- PHP follows project tabbing + WP coding standards. JSDoc on `<PersistentRegion>` reads clean.
- `data-app-mounted` DS-neutral data attr — respects kernel-DS-neutrality rule. ✓
- No new external deps. ✓

## Recommendation

Block until JS-side slot resolution actually mounts secondary apps + browser-smoke #2 checked. Unit-test surface gives false confidence — landing as-is ships dead feature.

dabowman:
Review addressed in \`a40334e\`. Blocker + all secondaries landed.

### Blocker — synthesized routes reachable from JS

Fixed via reviewer Option 1. Two-part change:

1. **\`PATTERN_VALID_RE\` widened** to accept \`@<slot>/<rest>\` shape: \`/^(@[a-z][a-z0-9-]*)?\/[A-Za-z0-9_/{}\-*]*$/\`. \`isValidRoutePattern('@detail/split')\` now returns true. Negative tests confirm \`@/posts\` (bare @) and \`@Detail/split\` (uppercase) still reject.

2. **\`routing.mode\` field added** to region declarations. Two modes:
   - \`query\` (default) — preserves existing palette behavior; slot value comes from URL query param of the same name (\`?palette=open\` → "open").
   - \`mirror\` — slot value synthesized as \`@<route-key><primary>\` (e.g. URL \`#/split\` + key \`detail\` → \`@detail/split\`). Engine peer regions opt in; the compiler's \`@<slot>/<primary>\` routes become findable without polluting the URL.

\`core:default\` engine \`detail\` region declares \`mode: "mirror"\`. Existing palette flow unchanged (no mode declared → default query mode).

### Secondary fixes

- **\`$primary_path\` duplication consolidated.** Single \`$primary_route_key\` derived once per screen, shared by all three synthesis branches (\`_self\`, slot-routed-screen, multi-app peer).
- **\`_self\` defensive guard tested.** Non-primary entry with \`slot: "_self"\` now has explicit assertion: primary route stays on first \`apps[]\` entry; no \`@_self/<path>\` clobber.
- **Intra-slot collision tested.** Two non-primary entries claiming same slot → first-wins per existing \`! isset\` guard. Contract locked.
- **Parametric primary path tested.** \`/posts/{id}/edit\` with \`slot: "detail"\` peer. Slot route key inherits the \`{id}\` segment verbatim; per-entry config preserves \`{id}\` substitution markers (interpolate resolves at match time).
- **CSS rule generalized** from \`[data-region-id="detail"][data-app-mounted="false"]\` to \`[data-app-mounted="false"]:not([data-keep-visible-when-empty="true"])\`. Adding peer slots (inspector / preview / future) no longer needs parallel CSS updates. Opt-out attribute for regions that want to stay visible when empty.
- **\`data-app-mounted\` gated to routable regions.** Kernel emits the attr ONLY on regions declaring \`routing.route-key\` (not sidebar wrappers / container regions with child regions but no own app). Generic CSS rule can't accidentally collapse them.
- **Roadmap header** dropped "may be deferred" parenthetical from the ✅ Shipped line.

### Test deltas (all green)

- Node \`tests/runtime/match-route.test.mjs\`: 39 → **50** (+11 covering \`readSlot\` mirror mode + \`isValidRoutePattern\` \`@<slot>/<path>\` shapes).
- PHP \`run-v3-compiler-tests.php\`: 79 → **85** (+6: \`_self\` guard, intra-slot collision, parametric primary path × 3).
- PHP \`run-shape-tests.php\`: 128 (unchanged).
- PHP \`run-engine-defaults-tests.php\`: 22 (unchanged).
- \`npm run test:schema\`: 101.
- \`npm run test:runtime\` all green.
- Lint + build clean.

Browser-smoke #2 still pending — split-pane shell author + visual verification needs human eyes. But the synthesized route is now reachable from the JS layer (matcher accepts the pattern + mirror-mode readSlot returns the synthesized value) so the path is no longer dead at the runtime layer.

PR ready for re-review.

dabowman (commented):
## Re-review — commit `a40334e`

Prior blocker resolved. JS-side slot resolution now works end-to-end:
- `PATTERN_VALID_RE` widened to `^(@[a-z][a-z0-9-]*)?\/[A-Za-z0-9_/{}\-*]*$` — accepts `@<slot>/<rest>` (lowercase slot, hyphens). Verified via 5 new pattern-validity assertions.
- `readSlot(url, key, mode='mirror')` synthesizes `@<key>/<primary>` — verified URL `#/split` + `route-key: "detail"` + `mode: "mirror"` produces `@detail/split`, which matches the compiler's emitted route.
- `core:default` engine's `detail` region declares `routing: { route-key: "detail", mode: "mirror" }`. End-to-end flow correct.

Secondary fixes all landed:
- `$primary_route_key` consolidation (single source of truth across `_self` / slot-screen / multi-app peer branches).
- `_self` defensive-guard test added (2 assertions: primary route preserved + no `@_self/<path>` route emitted).
- Intra-slot collision test added (first-wins contract locked).
- Parametric primary path test added (`/posts/{id}/edit` × `slot: "detail"`; 3 assertions covering key emission + param preservation in config).
- `data-app-mounted` now scoped to routable regions only (won't accidentally collapse non-routable containers).
- CSS rule generalized to `[data-app-mounted="false"]:not([data-keep-visible-when-empty="true"])` — scales to new peer slots without parallel CSS update.
- Roadmap "may be deferred" phrasing dropped.

Test runs on tip (`a40334e`):
- `run-v3-compiler-tests.php` — PASS 85 (was 79, +6 per commit message).
- `match-route.test.mjs` — PASS 50 (was 39, +11).
- `npm run test:schema` — PASS 101.

## 🚨 New blocker — `admin-v3.json` schema rejects `mode` on author regions

`docs/schemas/admin-v3.json:820` — author-level `routing` block declares `additionalProperties: false` with only `route-key` defined. A workspace that authors its own `regions` block (v2 escape hatch per the description on line 818) and declares `routing: { route-key: "detail", mode: "mirror" }` will fail schema validation.

Schema tests pass today because:
- `admin-engine-v3.json:377` has `additionalProperties: true` on `routing` — the bundled `core:default/engine.json` is permitted.
- No fixture exercises an authored-workspace region with `mode`.

Fixes needed:
1. **`admin-v3.json`** — add `mode` property to `routing`:
   ```json
   "mode": {
       "enum": ["query", "mirror"],
       "description": "Slot resolution mode. `query` (default) reads the URL query parameter of the same name. `mirror` synthesizes `@<route-key>/<primary>` from the URL primary path — pairs with v3 compiler's `@<slot>/<primary>` route synthesis. See spec §6.4."
   }
   ```
2. **`admin-v3.json:826`** — `route-key` description still says "reads the same-named URL query parameter" unconditionally. Update to acknowledge `mode` selects between query + mirror.
3. **`admin-engine-v3.json:375`** — formal `mode` property declaration (currently passes via `additionalProperties: true`, but IDE feedback + parity with admin-v3 warrant declaring it explicitly).
4. **Schema fixture** — add an authored-workspace region declaring `routing.mode: "mirror"` to `tests/schema/validate-shells.test.mjs` fixture set, locking the schema accepts it.

## Polish

- Spec doc `docs/wp-admin-shell-design-spec.md` §5.4 / §6.4 — the new `routing.mode` field should be documented alongside `route-key`. Both schemas reference these sections; current spec prose predates the field.
- Roadmap §3c.4 implementation paragraph describes the original synthesis approach but doesn't mention the `mode: "mirror"` runtime contract. Worth a sentence so readers tracing the feature aren't surprised by the engine.json declaration.

## Recommendation

Address the schema gap (1–4 above) + spec doc note before merging. Mirror mode is the contract author-level regions will use to opt into multi-app screens — leaving it schema-rejected at the workspace layer ships a half-wired feature.

dabowman:
Re-review addressed in \`7b9bd22\`. Visual regression + schema gap both fixed.

### Visual regression — workspace broken / apps hidden

Root cause: the generic CSS rule (\`[data-app-mounted="false"]\`) caught the \`_self\` content region during transient empty-URL states (initial mount, hash navigation lulls). Content briefly disappeared → workspace looked broken.

**Fix:** \`<PersistentRegion>\` now emits \`data-app-mounted\` ONLY when \`region.routing.mode === 'mirror'\`. Previously was emitted on any region with \`routing.route-key\` (which included \`_self\` content + query-mode palette).

Collapse-when-empty is now exclusively the multi-app peer-slot contract. Engine peer regions (\`detail\`) opt in via \`routing.mode: "mirror"\`. Everything else (`_self\` content, query-mode palette, non-routable wrappers) keeps existing always-rendered behavior. Adding future peer slots (\`inspector\`, \`preview\`) inherits the contract via the mode declaration alone — no CSS or attr-emission updates needed.

### Schema blocker — author regions

Fixed all 4 items:

1. **\`admin-v3.json#/$defs/regionInstance.routing\`** — added \`mode\` property with \`enum: ["query", "mirror"]\` + full description.
2. **\`admin-v3.json\`** — \`route-key\` description updated to acknowledge \`mode\` selects the resolution strategy.
3. **\`admin-engine-v3.json\`** — declared \`mode\` explicitly (previously passing via \`additionalProperties: true\`).
4. **Schema fixtures** — \`02-author-region-routing-mode.json\` (positive, author workspace declaring \`mode: "mirror"\` on multiple regions) + \`06-bad-routing-mode.json\` (negative, \`mode: "reflect"\` rejected).

### Polish

- **Spec doc §5.4** — \`routing.mode\` documented alongside \`route-key\`. Mirror-mode collapse-when-empty contract spelled out. Link example updated to show the difference between query-mode (\`?detail=/posts/42\`) and mirror-mode (\`#/split\`) URL shapes.
- **Roadmap §3c.4** — runtime contract paragraph added describing the \`mode\` field + collapse behavior.

### Test deltas

- \`npm run test:schema\`: 101 → **103** (+2 new fixtures).
- \`npm run test:runtime\`: all green (no count change — match-route already covers \`readSlot\` mirror mode from prior commit).
- All PHP suites green.
- Lint + build clean.

PR ready for re-review. Browser-smoke regression should be resolved — content + palette + sidebar all back to always-rendered behavior; only mirror-mode peer regions (the new `detail` opt-in) participate in collapse-when-empty.

dabowman:
Visual regression fixed in \`dd7fa85\`. Two coupled bugs.

### Issue 1 — inline-style beat CSS \`display: none\`

\`core:detail\` template's \`default-style\` ships \`display: flex\` (plus \`inline-size: min(720px, 50%)\`, \`block-size: 100%\`). The kernel applies template default-style as INLINE \`style="..."\` on the region wrapper. The mirror-mode collapse rule lost to inline-style specificity — \`display: flex\` (inline) outranks \`display: none\` (stylesheet).

Result: detail region stayed visible + sized despite no app mounting, hijacking 50% of the viewport.

**Fix:** \`display: none !important\` in the engine collapse rule. Inline styles fall to \`!important\` stylesheet declarations. Surgical override, scoped to the data-attribute-gated selector.

### Issue 2 — detail rendered as straggler outside \`.wp-admin-shell-layout__body\`

\`Layout.js\` \`SLOT_IDS\` map only covered \`toolbar / sidebar / content / preview\`. Detail wasn't named → landed in the \`stragglers\` array → rendered as a direct child of \`.wp-admin-shell-layout\` (flex column), sibling to \`.wp-admin-shell-layout__body\`. With \`block-size: 100%\` it claimed the full vertical extent, pushing body into a tiny strip; sidebar compressed to its min-inline-size in the top-left of the shrunken body.

**Fix:** Add \`detail\` to \`SLOT_IDS\`. Render inside \`.wp-admin-shell-areas\` next to content + preview. When detail is present, container class flips to \`has-detail\` so existing flex arrangement extends naturally.

### Together

- Detail no longer disrupts layout (correctly slotted inside the body container).
- Collapses cleanly when no route matches (\`!important\` override).
- Sidebar + content keep their existing geometry.

### Test deltas

No new tests (this is a layout/CSS regression fix that depends on visual verification). All existing suites green:
- \`npm run test:schema\`: 103.
- \`npm run test:runtime\`: all green.
- \`npm run lint:js\` / \`lint:ts\`: clean.
- \`npm run build\`: clean.

Pull \`dd7fa85\` + reload — sidebar + content should be back to full geometry, detail region absent from view (collapsed via \`display: none !important\` since no route matches).

dabowman (commented):
## Re-review #2 — tip `dd7fa85` (commits `7b9bd22` + `dd7fa85`)

Prior schema-gap blocker resolved + bonus visual regression caught + fixed.

### Schema gap — fully addressed
- `admin-v3.json:826` — `mode` property added with `enum: ["query", "mirror"]` + full description covering mirror semantics + collapse-when-empty + engine peer-region opt-in pattern.
- `admin-v3.json:822` — `route-key` description rewritten; now says `mode` selects resolution strategy (query default vs mirror) instead of declaring "any other value reads URL query parameter" unconditionally.
- `admin-engine-v3.json:380` — `mode` declared formally (parity with admin-v3, even though `additionalProperties: true` permits it).
- Fixtures: `positive/02-author-region-routing-mode.json` exercises `_self` + `mirror` (detail+inspector) + `query` (palette) variants; `negative/06-bad-routing-mode.json` rejects `mode: "reflect"`. Schema sweep 101 → 103.
- `docs/wp-admin-shell-design-spec.md` — +11/-3, picks up `mode` field.

### Visual regression — additionally fixed in `dd7fa85`
- **`data-app-mounted` scope narrowed.** Prior commit `a40334e` emitted on every `routing.route-key` region — caught `_self` content region during empty/transient URLs → content disappeared. Now gated to `routing.mode === "mirror"` only. Multi-app peer-slot contract honored, `_self` + query-mode regions keep their always-rendered behavior.
- **Inline-style override.** `core:detail` template's `default-style: { display: flex, inline-size: min(720px, 50%), … }` is applied as **inline** `style="…"` on region wrapper by kernel. Collapse rule `display: none` lost to inline specificity. Fix: `display: none !important` in engine CSS. Comment justifies the override. Acceptable surgical fix — inline default-style is a kernel-applied attribute the engine CSS can't easily reach.
- **Detail slot integration in Layout.js.** Added `detail` to `SLOT_IDS` map + `findById` + `claimed` set + rendered inside `.wp-admin-shell-areas` container next to content/preview. Without it the region was rendering as a straggler outside `__body`, claiming full vertical extent + crushing sidebar/content to a corner strip.

### Test runs on tip
- `npm run test:schema` → PASS 103 (was 101, +2 fixtures).
- `npm run test:runtime` → PASS 27/27 totals.
- `run-v3-compiler-tests.php` → PASS 85 (unchanged since `a40334e`).

### Polish — non-blocking nits
- **Dead className markers.** `Layout.js` emits `has-detail` (and pre-existing `has-preview`) on `.wp-admin-shell-areas`, but no CSS rule consumes either class. Width math is entirely flex-distributed via each child's inline `default-style`. Classes are harmless forward-looking markers — either drop them or add rules that flip behavior (e.g., `gap`, column count) so the markers earn their keep.
- **`!important` is a smell long-term.** The root cause is the kernel emitting template `default-style` as inline style, which forces engine CSS into `!important` whenever it needs to override. Future-direction: a template field like `default-style.collapsible: true` so kernel skips emitting `display: flex` inline when `data-app-mounted="false"`, letting engine CSS win without `!important`. Out of scope for this PR — note for the v3 polish queue.
- **`mode` enum choice.** `"mirror"` reads well in code but is slightly opaque vocabulary. Alternatives like `"path"` or `"primary"` describe the source more directly. Bikeshed — fine to keep.

### Recommendation

Approve in spirit. All my prior blockers addressed, regression caught + fixed mid-review cycle, schema + fixtures + spec docs all touched. The `!important` + dead `has-detail` class are technical-debt notes worth filing as follow-ups but don't block this merge.

PR #59, opened 2026-05-21
v3: 3d.4 — documentation sweep

dabowman:
## Summary

Phase 3d.4 of `docs/v3/roadmap.md`. Sweeps CLAUDE.md, master spec, schema-sketch, public reference docs, and roadmap so v3 is the canonical surface; v2 references are historical-only.

## What changed

| File | Edit |
|---|---|
| `CLAUDE.md` | Status block rewritten (phases 3a–3d.5 enumerated; v1/v2 marked historical). Pre-reads reordered v3-first. Test totals refreshed (~1300 assertions). Project structure tree updated with new PHP cascade classes + runtime modules. Navigation + multi-area sections rewritten around v3 vocabulary. |
| `docs/wp-admin-shell-design-spec.md` | v3 reshape header; §4.3 v3 admonition; §5 documents v3 region declaration sources; §6.2 routes block clarified as runtime-internal w/ v3 author-shape example; §14 gains v2 → v3 migration paragraph; §15 v3 status header; §19 references surface v3 docs first. |
| `docs/v3/schema-sketch.md` | Reframed "Working draft" → "Canonical v3 schema design doc"; bidirectional cross-link with master spec; open-questions refreshed (variant URL routing marked resolved by dataview-registry restoration). |
| `docs/public/admin-json-reference.md` | Full rewrite around v3 shape: workspace/settings/screens/menu/commands top-level blocks; v2-surfaces-deprecated table at end; v3 schemas referenced everywhere. |
| `docs/public/app-json-reference.md` | viewConfig → dataView (variants family + extends); new slots / slotHints sections; `$schema` URL bumped to v3. |
| `docs/public/engine-json-reference.md` | New modes / slots / menu-renderer / defaultRegions sections; `$schema` URL bumped to v3; required-fields list updated. |
| `docs/v3/roadmap.md` | 3d.4 ✅ Shipped; phase status table refreshed; 3d.3 remains the only open Phase 3d item. |

5 commits, 7 files, +668 / −338 lines.

## Schema-sketch fate

Promoted in place rather than consolidated into spec. 898 lines of v3-specific schema design rationale would have ballooned the spec; bidirectional cross-link makes it clear which doc is authoritative for what (spec = normative; sketch = design rationale). Sketch now reads as a canonical companion doc, not a draft.

## Intentionally left as-is

- v2 admin.json example in spec §4.3 — still the runtime-internal shape post v3-compiler synthesis; useful for region declaration reasoning. v3 admonition added pointing at schema-sketch.
[...]

Issue #30, opened 2026-04-30
Runtime smoke harness: kernel mount + content-region routing + nav prune (JSDOM)


dabowman:
Cover the bug class neither schema validation nor the resolver-shape / synthesis tests catch: React component-level regressions where the resolved config is correct but a runtime reader mounts the wrong app, routes to the wrong path, or drops a nav item.

`tests/runtime/build-runtime-config.test.mjs` now covers the kernel's region / route / default-route / command **synthesis** from the v3 shape, but nothing mounts the kernel and asserts on rendered output.

## Scope

Headless mount of the kernel via JSDOM (or Playwright if a real browser becomes warranted) against fixture configs. Assertions:

- `kernel(config)` returns a React tree without throwing.
- The primary content region mounts the expected app for `workspace.default-screen` (e.g. a screen with `app: "core:posts"` ⇒ PostsApp rendered).
- `core:navigation` renders ≥ 1 item for a non-empty `menu` tree.
- `core:command-palette` registers the expected "Go to {screen}" commands against `@wordpress/commands`.
- Capability gating: a fixture with role-equivalent missing caps drops the right screens + nav items (mirror the role matrix from `tests/php/run-cap-gating-smoke.php`).
- Token emission reaches the rendered DOM through the active engine's `ThemeProvider`.

dabowman:
@claude validate and address in a new PR

PR #60, opened 2026-05-21
v3: 3d.3 — test surface rewrites

dabowman:
## Summary

Phase 3d.3 of `docs/v3/roadmap.md`. Ports cap-gating-smoke to v3 (was SKIP), modernizes shape-tests, adds multi-app coverage, fixes two latent bugs surfaced by the smoke port.

## What changed

| File | Edit |
|---|---|
| `tests/php/run-cap-gating-smoke.php` | Rewritten end-to-end. Was SKIP (5 → **38** assertions). Exercises v3 screen-level + menu-level capability gating across roles. Covers OR-semantic `permissions.capabilities[]` + `permissions.roles[]` + trust-tier cascade rules (role/user origin grants restrict-only). |
| `tests/php/run-shape-tests.php` | v1 branch dropped; v3 assertions + multi-app slot-route walk added. 119 → **133** (+14). |
| `wp-admin-shell.php` | **Latent bug #1 fix:** added `require_once class-wp-admin-shell-permissions.php` (defined-but-never-loaded since 3b). **Latent bug #2 fix:** extended `wp_admin_shell_resolve_capabilities()` to walk `screens[].permissions.capabilities[]`. `run-cap-tests.php` had been silently failing 17/54 since 3d.1 dropped v2 default; now 54/54. |
| `CLAUDE.md` | Test surface paragraph + Testing-section count refresh. |
| `docs/v3/roadmap.md` | 3d.3 marked ✅ Shipped with landing notes; status table updated. |

## Latent bugs fixed as side-effects

1. **`WP_Admin_Shell_Permissions` never required.** Shipped in Phase 3b but the plugin bootstrap missed it. Defined-but-never-loaded. Cap-gating-smoke port required loading it; included that require in the same commit.
2. **`wp_admin_shell_resolve_capabilities()` skipped v3 screen perms.** Walked only v2 region.capability fields. The runtime cap-map shipped via `wp_add_inline_script` was missing every v3 cap. `run-cap-tests.php` silently failing 17/54 since v3 default landed in 3d.1; now 54/54.

## Per-suite assertion counts (final)

**PHP — 885 total:**

| Suite | Count | Δ |
|---|---|---|
| run-cascade-tests | 39 | — |
| run-manifest-tests | 67 | — |
| run-cap-tests | **54** | +17 (was silently failing) |
| run-shape-tests | **133** | +14 |
| run-tokens-tests | 13 | — |
| run-engine-defaults-tests | 22 | — |
| **run-cap-gating-smoke** | **38** | **+33 (was SKIP)** |
| run-chromeless-bridge-tests | 13 | — |
| run-data-view-tests | 106 | — |
| run-mode-resolution-tests | 20 | — |
| run-v3-compiler-tests | 85 | — |
| run-migrate-shell-cli-tests | 81 | — |
| run-classic-menu-bridge-tests | 66 | — |
| run-preload-tests | 22 | — |
| run-menu-route-shims-tests | 70 | — |
| run-dashboard-widgets-tests | 56 | — |

**Node — 694 total:** 510 runtime + 103 schema + 4 parity + 77 TS engine.

**Grand total: 1579 assertions, all green.** Roadmap target was ~1000-1200; landed above.

## Dropped assertions

**Zero.** Every audited v2-only branch turned out to be either compiler-back-compat translator coverage (still needed) or runtime primitive coverage (bindings parser, triggerStore) still in use by v3-shape code paths.

## Surfaced gaps (not closed)

- **React `<PersistentRegion>` `data-app-mounted` attribute** exercised end-to-end via engine CSS collapse + bundled shells, but lacks a unit-level JSDOM assertion. Tracked under existing JSDOM-scaffold issue #30.

## Conflict note

This PR touches `CLAUDE.md` + `docs/v3/roadmap.md` — same files as open PR #59 (3d.4 doc sweep). Merge order doesn't matter; the loser will rebase + resolve trivially. Both PRs touch separate sections of each file.

## Test plan

- [x] All PHP suites green via `wp eval-file`.
- [x] `npm run test:schema` PASS 103.
- [x] `npm run test:runtime` PASS 510.
- [x] `npm run test:parity` PASS 4.
- [x] `npm run test:engines` PASS 77.
- [x] `npm run lint:js` / `lint:ts` clean.
- [x] `npm run build` clean.

## After merge

**Phase 3d complete.** All 5 sub-phases shipped (3d.1 / 3d.2 / 3d.3 / 3d.4 / 3d.5).


dabowman (commented):
# Review — `v3: 3d.3 — test surface rewrites`

## Overview

Phase 3d.3 of the v3 roadmap. Three things land together:

1. **`run-cap-gating-smoke.php` rewritten end-to-end** (5 → 38 assertions). Replaces the v2 nav-pruning smoke (the v2 region tree it walked was retired in 3d.1) with v3 screen + menu capability gating via `WP_Admin_Shell_Permissions`: OR-semantic cap/role evaluation, app-floor AND backstop, unknown-slug fail-closed, super-admin magic, trust-tier shrink-only with audit log, and an end-to-end role-walk over `wp-admin-default` asserting monotonic visibility.
2. **`run-shape-tests.php` modernized for v3** (119 → 133). Dead v1 branch dropped; adds top-level `workspace`/`screens` presence + multi-app `@<slot>/<path>` route synthesis walk.
3. **Two latent bug fixes uncovered while porting:**
   - `wp-admin-shell.php`: missing `require_once …/class-wp-admin-shell-permissions.php` (file existed since Phase 3b, never loaded by bootstrap).
   - `wp_admin_shell_resolve_capabilities()` extended to walk `screens[].permissions.capabilities[]` — the v3 cap-precompute collector was dropping every v3-declared cap, silently failing `run-cap-tests.php` 17/54 since 3d.1.

## Code quality & correctness

### Strong points

- The end-to-end **monotonic-visibility** invariant (`subscriber ≤ contributor ≤ author ≤ editor ≤ admin`) is the right shape: it asserts the *property* the permission cascade must satisfy rather than coupling tests to authoring decisions on exact screen ids. Survives screen-catalog churn.
- Trust-tier coverage is thorough: shrink-allowed case, grow-rejected case, audit-entry shape (`origin` + `kind` + `attempted`), and consumer-can-shrink-cap-AND-role together. Matches `WP_Admin_Shell_Permissions::enforce_trust_tiers` semantics exactly.
- `resolve()` defaults are tested across **three** distinct inputs (`null`, `{caps:[], roles:[]}`, well-formed) — locks in the "fail-closed by convention" rule from the class docblock.
- Shape-test multi-app slot-route walk catches a real class of regression: v3 compiler `synthesize_routes()` drift between authoring (`screens[].apps[].slot`) and synthesized `@<slot>/<path>` keys.
- The bootstrap fix is a one-liner with a clear comment trail in the PR body; the cap-precompute extension is the minimum-viable walk that mirrors the existing region.capability walk pattern.

### Concerns / suggestions

- **Smoke test depends on environment-provided users.** `wpas_cap_smoke_user_for_role()` looks up users by login → fallback to first user in role → null. If wp-env has only an admin fixture, the e2e section silently skips four out of five roles and the monotonic-visibility assertion becomes vacuous (admin-only, single data point, trivially monotonic). The neighbouring `tests/php/run-cap-tests.php` already solves this with `ensure_user()` — recommend porting that pattern (create per-role users at the top of the smoke, delete in a register_shutdown_function). At minimum, assert at least three roles were actually walked, otherwise the suite would silently regress to "covers admin only".

- **Menu pruning section resolves once as admin, prunes per role** (around `tests/php/run-cap-gating-smoke.php:502-538`). Works for `wp-admin-default` because no role/user origin overrides the menu shape there, but it's a hidden assumption — a future shell that does customize the menu per role would silently make this smoke wrong (admin's menu tree pruned by editor caps ≠ editor's actually-resolved menu tree). Worth a one-line comment noting the assumption, or matching the per-role re-resolve pattern of the screen-visibility walk above it.

- **`wp_admin_shell_resolve_capabilities()` walks screens but not menu items.** v3 menu items can carry their own `permissions.capabilities[]` (the smoke test even prunes the menu by per-item screen-binding caps); if a menu item has an inline cap with no bound screen, the precompute walk misses it and `userCan()` in the runtime cap-map will get a default-false. Tracking under the existing latent-bug pattern would justify a third walk here; if intentional (because menu caps are always also on the bound screen), worth a comment to that effect on the new block.

- **Shape-test fallback path drifts slightly from compiler.** Test builds `$path = /screen_id` when `screen.path` is empty. Compiler (`class-wp-admin-shell-v3-compiler.php:629-637`) early-continues the entire screen synthesis when `path === '' && slot === '_self'`, never reaching multi-app peer synthesis. In practice all bundled shells supply explicit paths so this never bites, but a path-less primary with a slotted peer would make the test assert a route the compiler intentionally skipped. Either mirror the compiler's early-continue in the test or note "bundled shells all carry paths" inline.

- **`grid` slot exemption is a behavioural smell, not a real exemption.** The test hardcodes `if ( $entry_slot === 'grid' ) continue;` with a comment that says "shouldn't emit a slot-namespaced route". The compiler actually *does* emit `@grid/<path>` routes — the slot is just app-internal at the runtime layer. The exemption is harmless but the comment misleads; suggest "skip — synthesized but ignored by the runtime (app-internal composition)" so the next reader doesn't try to enforce non-emission in the compiler.

- **Repeated `WP_Admin_Shell_Cache::flush()` in the role-walk loop** rebuilds the entire cascade five times. Negligible at current scale but worth noting if the resolver gets heavier; `reset_request_memo()` alone may suffice when only `wp_set_current_user()` changed between iterations (user/role origins are the only inputs that vary). Not blocking.

- **Monotonic loop initializes `$prev_role = ''`**. On a non-monotonic first iteration the echo would read `role  had -1 visible, X has Y` — cosmetic, low-priority.

## Project-convention adherence

- Test file naming + invocation header match the suite convention.
- Bug fix in `wp-admin-shell.php` keeps the existing v0/v1/v2 walk layering and adds v3 as a sibling block, mirroring the comment style of the surrounding code.
- CLAUDE.md "Test surface" paragraph updated with the new counts + explanatory parentheticals.
- Roadmap entry updated with the actual landings rather than a forward-looking plan.
- Conflict note in the PR body flags the CLAUDE.md / `docs/v3/roadmap.md` overlap with PR #59 — coordination needed but trivial.

## Test coverage

- The PR is *itself* the test-coverage delta. 5 → 38 for cap-gating-smoke, +14 for shape, +17 cap-tests (restoration). Total claimed: 1579.
- One known gap explicitly flagged: `<PersistentRegion>`'s `data-app-mounted` attribute lacks a JSDOM unit-level assertion (tracked under issue #30).
- Suggested follow-up gaps from this review: per-role user fixture creation in cap-gating-smoke; possibly menu-item perms walk in `wp_admin_shell_resolve_capabilities()`.

## Security

- Trust-tier audit log is the right shape (origin / kind / attempted / detail). Tests confirm the role/user-origin grow rejection path so the security invariant is locked in.
- Unknown-slug fail-closed is asserted for both caps and roles, matching the class docblock contract.
- No new attack surface; the bug fix in `wp_admin_shell_resolve_capabilities()` adds caps to a *precomputed boolean map* (advisory hint for the runtime); the authoritative cap check still goes through WP's `current_user_can()` server-side.

## Recommendation

**Approve with minor follow-ups.** Solid PR — well-scoped, both bug fixes are surgical with full justification, the test redesigns are property-based rather than fixture-coupled, and the monotonic-visibility assertion in particular is a nice invariant that will catch real regressions cheaply. The main thing worth addressing before merge is creating per-role test users in the smoke (so the e2e assertions don't silently degrade to admin-only when wp-env is sparsely populated); the other items are clean-up suggestions for a follow-up commit.

---

PR #61, opened 2026-05-22
v3: PR #49 review — security cascade hardening (Stage 1)

dabowman:
## Summary

Stage 1 of the [5-PR plan](docs/plans/2026-05-22-pr49-pre-merge-feedback.md) landing PR #49 pre-merge review feedback. Closes 5 of 7 🔴 blockers + 1 🟠 major + folds Stage 2 (`user-switchable` kebab fix).

## Items addressed (from PR #49 review)

- **🔴 Blocker — Trust-tier enforcement dead code.** `WP_Admin_Shell_Permissions::enforce_trust_tiers()` had zero callers. Wired in via new companion `enforce_origin_tier()` that fits the resolver's per-origin loop. `core` / `engine` / `plugin` / `site` add+remove permissions; `role` / `user` shrink-only.
- **🔴 Blocker — Null-tombstone bypass.** `merge.php` lines 108-111 (assoc-null) + 205-212 (`__tombstone: true`) fired regardless of authority. Split `$authoritative` → separate `$tombstone_authority`; new `merge_with_tombstones()` for site origin. Consumer-origin null/tombstone payloads now no-op with `WP_DEBUG` notice.
- **🔴 Blocker — `customizable` pass-through.** Replaced verbatim pass-through on v3 top-level blocks (`menu` / `screens` / `commands` / `workspace` / `preload` / `regions` / `routes`) with per-field path-allowlist enforcement (v2 semantics). Default-deny at consumer origins; trust tiers bypass. Hardcoded deny-list (`screens.*.permissions`, `screens.*.app`, `commands.*.invoke`, `workspace.engine`) blocks even matching allowlist entries.
- **🔴 Blocker — `is_safe_href` protocol-relative bypass.** `//evil.example.com` now rejected.
- **🟠 Major — REST `/data-view` over-discloses.** `permission_check()` was `is_user_logged_in()` only. Screen-keyed requests now gate against requested screen's resolved `permissions` block via `WP_Admin_Shell_Permissions::user_passes()`. 403/404/401 paths covered. Triple-keyed lookups keep logged-in floor (not screen-scoped).
- **🔴 Blocker (folded from Stage 2) — `user-switchable` key mismatch.** Code read camelCase `userSwitchable`; schemas + shells declare kebab `user-switchable`. Code → kebab everywhere (incl. `class-wp-admin-shell-config.php`, `class-wp-admin-shell-cli.php`).
- **Filter contract docblock** — `wp_admin_shell_data` filter docblock added at resolver dispatch capturing "receives author shape, not compiled shape; per-origin filters at priority 5 are the supported route synthesis path."

## Locked decisions honored

- `customizable` scope: **per-field** path-allowlist
- `user-switchable` direction: **code → kebab**
- Hardcoded deny-list: 4 paths never allowlistable from consumer origins
- Trust tier: `core`/`engine`/`plugin`/`site` may add+remove; `role`/`user` shrink-only

## Test surface delta

**New PHP runners (+45 assertions):**
- `tests/php/run-security-cascade-tests.php` — 35 assertions covering trust-tier enforcement audit entries; null-tombstone three-way split (`merge` / `merge_with_tombstones` / `merge_authoritative`); `customizable` per-field allowlist + hardcoded-deny + trust-tier bypass; `is_safe_href` protocol-relative end-to-end via `wp_admin_shell_register_menu_item`.
- `tests/php/run-data-view-rest-tests.php` — 10 assertions covering subscriber → admin-only 403, admin → 200, unknown screen 404, triple-keyed `?kind=X&name=Y` keeps logged-in floor, logged-out 401.

**Modified runner:** `tests/php/run-cascade-tests.php` (+1 assertion: `user-switchable` kebab form via `desktop-demo.v3.json` fixture).

**Merge gate (all green):**
- 19 PHP runners: 932 assertions
- `npm run test:schema` / `test:parity` / `test:runtime` / `test:engines` — no JS surface in this PR; confirmed no regressions
- `npm run lint:js` clean

## Test plan

- [ ] Cap gating walkthrough on `wp-admin-default.json` as subscriber → editor → admin → super-admin
- [ ] Attempt user-origin write to `screens[X].permissions` via REST `/user-prefs` — expect reject
- [ ] As subscriber, `GET /wp-admin-shell/v1/data-view?screen=plugins` — expect 403
- [ ] As admin, same request — expect 200 + valid doc
- [ ] Toolbar user-switching dropdown (admin) on `wp-admin-default.json` — expect functional (was always-false pre-PR)
- [ ] As `site` origin (REST `/wp/v2/settings`), write `{ "screens": { "comments": null } }` — comments screen drops
- [ ] As `user` origin, same payload — payload ignored, comments screen survives

## Plan deviations

- Split `$authoritative` flag into separate `$tombstone_authority` for cleaner site-origin semantics. Added `enforce_origin_tier()` companion method alongside existing `enforce_trust_tiers()` — per-origin incremental call fits resolver loop naturally; both APIs coexist + both test-covered.
- Audit: zero bundled shells declare `customizable` blocks for v3 top-level — no in-tree breakage from enforcement.


dabowman:
## Code Review — PR #61 security cascade hardening

### Overview

Five security blockers + one major + the user-switchable kebab fix landed in one coherent PR, executed largely faithfully to the locked Stage-1 plan. The trust-tier wire-in (`enforce_origin_tier`), null-tombstone gating via the new `merge_with_tombstones()`, hardcoded customizable deny-list, `is_safe_href` protocol-relative reject, and REST `/data-view` screen-keyed gating are all real fixes that close the documented attack paths. Test coverage on the happy paths is solid (37 + 11 assertions, plus the kebab fixture in cascade). However, I found **one merge-shape correctness bug**, **one missed JS callsite for the user-switchable rename**, and a handful of `is_safe_href` bypass variants the validator still admits. None of the remaining items reopen the v3 security gates; the most consequential is the list-form shape corruption in `filter_v3_block`. **Not merge-blocking** on the security thesis — security model holds — but the list-shape bug should land before any consumer-origin shell starts using `customizable` on `commands`/`preload`.

### 🔴 Blockers

None on the security model itself — the trust-tier model is sound, the cascade gating is correctly ordered (customizable → enforce_origin_tier → merge), and defense-in-depth holds (deny-list + shrink-only both fire). All five claimed blocker fixes work for the documented attack surface.

### 🟠 Major

- **`filter_v3_block` corrupts list-form blocks on consumer-origin survival.** `class-wp-admin-shell-customizable.php:225-235`. `collect_leaves` flattens a list-of-keyed-objects (e.g. `commands: [{id:'save', shortcut:'Mod+S'}]`) by id, producing dot-paths like `commands.save.shortcut`. After deny-list + allowlist filtering, the rehydration at `dot_set( $out, $rel, $value )` reassembles into an **assoc map** (`{ save: { shortcut: '...' } }`). When `WP_Admin_Shell_Merge::merge()` then receives map-shape `commands` on top of list-shape base, the `base_is_assoc !== over_is_assoc` branch at `merge.php:103-106` triggers and the *entire base list is replaced* by the map. End state: `commands` block becomes a map at runtime, breaking `compileCommands.mjs` and any consumer expecting `[]`. Same bug latent for `preload` (integer-indexed survives roundtrip as an indexed map and would re-coerce; less catastrophic) and `routes`/`regions` if list-shaped overrides survive.

  No bundled shells exercise this today (per PR text + audit), but the moment a plugin author publishes a v3 site config with `commands` allowlisted via `customizable`, this corrupts shape silently. Repro test missing — the existing deny-list tests all assert `empty( $filtered['commands'] )`, which is the SENTINEL-drop path, not the survival path.

  Fix: track the original `$over` block shape (list vs map) in `filter_v3_block`; rebuild list-shaped blocks back into a list-of-objects keyed by id. Or, simpler, omit list-form blocks from per-field walker scope (whole-block replacement only when explicitly allowlisted at root). Either way, add a survival-path test for `commands.save.shortcut`.

- **`src/apps/user-menu/index.js:27` still reads `s.userSwitchable`** (camelCase) on the shells array. The PR claims "Code → kebab everywhere"; the JS surface escaped the rename. End-to-end this still works today because `wp-admin-shell.php:915` and `:925` both EMIT `userSwitchable` (camelCase) to JS via the read-both-emit-camelCase bridge in `wp_admin_shell_get_available_shells()`. So the toolbar dropdown DOES function after this PR — but the rename is incomplete: the schema-canonical key (`user-switchable`) never makes it across the PHP→JS boundary. Either flip the JS to `s['user-switchable']` and update the bridge to emit kebab, or document in a code comment that the JS contract intentionally diverges from schema casing. Right now it just looks like a missed grep.

### 🟡 Minor

- **`merge_keyed_arrays` non-authoritative `__tombstone: true` pollutes the merged entry.** `merge.php:259-275`. After the `_doing_it_wrong` notice fires for a consumer-origin tombstone, execution falls through to `if ( isset( $over_index[ $id ] ) ) { … merge_internal( $entry, $over_index[ $id ], … ) }`. `$over_index[$id]` carries `__tombstone => true`, which gets deep-merged into the surviving base entry. Result: base entry now has a stray `__tombstone: true` field stamped on it, propagated to JS, may confuse downstream consumers. Test at `run-security-cascade-tests.php:185-202` only checks the id survives, not the absence of `__tombstone`. Fix: after the WP_DEBUG notice, `unset( $over_index[$id]['__tombstone'] )` (or just `continue` to skip the merge entirely — falling through is the source of the pollution).

- **`is_safe_href` bypasses still admit several protocol-relative variants:**
  - **Leading whitespace.** `" //evil.example.com"` (one or more leading spaces/tabs) passes `strpos( $href, '//' ) === 0` (false — `'//'` is at position 1+) and falls through to `strpos( $href, ':' ) === false` → **accepted**. HTML5 mandates browsers strip leading ASCII whitespace from `href` before navigation, so the navigation target IS `//evil.example.com`. `class-wp-admin-shell-menu-items.php:484-512`. Fix: `$href = ltrim( $href )` before the protocol-relative check, or reject any whitespace-leading href.
  - **Backslash-protocol.** `"\\\\evil.example.com"` (two backslashes) — same fall-through, **accepted**. Most modern browsers don't navigate this as protocol-relative (per WHATWG), but Edge/IE legacy and some embedded WebViews historically do. Defensive reject is cheap.
  - **`vbscript:`** — claimed rejected in the docblock but the only test (`run-security-cascade-tests.php`) covers `javascript:` + `data:`. The current allowlist regex falls through on `vbscript:` (it does have `:`, isn't http/ftp/mailto/tel/sms, no leading `/`/`#`/`{` → returns false). So it IS rejected, just not asserted. Add one assertion to document the contract.

- **Test coverage gap — REST permission_check on a screen with no `permissions` block at all.** The plan called for "Subscriber `GET ?screen=profile` (no cap floor / `read` only) → 200". The current `profile` screen ships explicit `capabilities: ['read']`, so the assertion exercises the explicit-permissions path, not the implicit-default path. Either add a fixture with `permissions: null` / no block at all and verify it routes through `default_permissions()` (admin-only), or document in a test comment that this case is covered by run-cap-gating-smoke.

- **`enforce_origin_tier` has no integration coverage through `resolve_with()`.** Tests exercise the method directly with hand-rolled `$origin_doc`/`$merged_so_far`. The wire-in (`resolver.php:122`) isn't covered by an end-to-end test that loads a malicious user-origin doc and asserts the cascade strips the cap. Risk: a future refactor could drop the resolver call and unit tests would stay green. Add one fixture: site option + user meta containing a `screens.profile.permissions.capabilities: ['manage_options']` override; assert the resolved doc has it stripped.

- **`shrink_against_baseline` and the standalone `enforce_trust_tiers` have duplicated audit-emit logic.** `permissions.php:181-235` (in `shrink_against_baseline`) and `:301-369` (in `enforce_trust_tiers`) are near-identical error_log + audit-push paths. Extract a helper. Not blocking; just future-maintenance debt.

- **`array( 'capabilities' => array(), 'roles' => array() )` literal at `permissions.php:81-82`** — same shape inflates in three more places. Tiny constant or factory helper would dedupe.

### ⚪ Nits

- **Docblock claim "before customizable filtering" in resolver.php is imprecise.** The added comment at `resolver.php:121-141` says per-origin filters fire "before customizable filtering" — but customizable filtering only applies to CONSUMER origins (role/user), not to trusted plugin origins where most plugin authors hook. The wording could mislead a plugin author into thinking their `wp_admin_shell_data_plugin` hook runs through `Customizable::filter_doc`. Suggest: "before customizable filtering (for consumer origins) and before the merge".

- **`esc_html__()` + `esc_html()` double-escape** inside `_doing_it_wrong` calls (`merge.php:117, 270`). `_doing_it_wrong` already escapes at output time. Harmless but redundant.

- **`error_log` in `shrink_against_baseline`** vs `_doing_it_wrong` in `merge_internal`. Inconsistent within the same PR. The original PR #49 review-addendum called for canonical `_doing_it_wrong()` adoption; the plan-doc moved it Out of scope, so accepting either as fine — but mixing the two within Stage 1 is awkward.

- **`WP_Admin_Shell_Customizable::CONSUMER_ORIGINS = ['role', 'user']`** vs **`WP_Admin_Shell_Resolver::CONSUMER_ORIGINS = ['site', 'role', 'user']`**. Two different definitions of "consumer" with different semantics. Functionally correct (site is consumer to the resolver loop, trust-tier for permissions/customizable), but a future contributor will trip over this. Suggest renaming the resolver const to `MUTABLE_ORIGINS` or `CASCADE_TAIL_ORIGINS`.

- **`v3.0` version string in `_doing_it_wrong` calls.** v3 hasn't tagged a beta yet; using `v3.0` as the introduced-in version is forward-looking. Convention is to use the next ship version, so this is probably fine, but worth a sanity check at tag-time.

- **Test runner file headers** — `run-security-cascade-tests.php:18` and `run-data-view-rest-tests.php:19` invoke comments reference the file under `wp-content/plugins/WordPress-Admin-Environment/...`. Other PHP runners follow the same convention. Consistent.

- **CLI column rename** (`userSwitchable` → `user-switchable` in `wp admin-shell list` output) is technically a breaking change for any script scraping the column header. Probably nobody is scripting this, but worth a CHANGELOG line.

- **`assert_eq` for `is_wp_error( $result ) ? (int) $result->get_error_data()['status'] : null`** — clever but loses the `WP_Error` object's message context on assertion failure. A small helper `assert_wp_error_status( $label, $result, $expected_status )` would dump the full error message when status doesn't match. Future-maintenance.

Both claimed deviations are well-motivated and safe:

1. **`$authoritative` → `$tombstone_authority` split.** Clean refactor. The site origin's "additive but tombstone-honoring" semantics genuinely don't fit a single boolean. Three explicit entry points (`merge` / `merge_with_tombstones` / `merge_authoritative`) read clearer than one `merge_internal( …, true|false )`. Back-compat shim (`$tombstone_authority === null → (bool) $authoritative`) preserves the old `merge_with_restrict` alias. No subtle defaults — recursive calls always pass both args explicitly. **Approved.**

2. **`enforce_origin_tier()` companion alongside `enforce_trust_tiers()`.** Two methods with overlapping responsibility, both kept for legitimately different call shapes (per-origin loop vs. standalone per-origin map). The audit log is shared — both push to `self::$audit`. The DRY duplication noted above (Minor) is real but doesn't undermine the choice. Test coverage exists for the new method; the standalone one is unchanged. **Approved.**

Neither deviation introduces a subtle bug; both stay inside the locked decisions (trust-tier scope, customizable scope, kebab direction).

### What's right

- **Defense in depth on screens[].permissions.** Customizable deny-list (`screens.*.permissions`) and `enforce_origin_tier` shrink-only both fire on the same payload. Either layer catches a user-origin escalation attempt. Belt-and-suspenders here is the right call given the blast radius.
- **`merge_with_tombstones()` is the right primitive.** Site origin needed additive-on-arrays + tombstone-on-keys, which neither `merge()` nor `merge_authoritative()` supplied. New name is searchable.
- **REST `/data-view` permission gate distinguishes 401/403/404 correctly** (`is_user_logged_in` first → 401, then resolve screen → 404 if missing, then `user_passes` → 403 if denied). HTTP-status semantics match WP REST convention.
- **`sanitize_key` on `screen` param** + `sanitize_segment` on kind/name/variant. WP-canonical, no rolled-own validation. Good.
- **Test invocation lines + ABSPATH guards** present on both new runners. WP convention. Per-user fixturing via `ensure_user()` + cleanup via `register_shutdown_function`. Solid test hygiene.

1. **List-shape corruption in `filter_v3_block`** (🟠 Major). Add a survival-path test for `commands.save.shortcut` and confirm `commands` is still a list after the walker. Fix the rehydration to preserve list shape.
2. **`src/apps/user-menu/index.js:27` user-switchable** — either flip to `s['user-switchable']` + emit kebab from PHP, or add a code comment documenting the deliberate divergence. Currently looks like an incomplete grep.
3. **`is_safe_href` whitespace bypass** (🟡 Minor) — `ltrim( $href )` before the `//` check, plus one negative test for `" //evil.com"`.
4. **`__tombstone: true` shape pollution** on consumer-origin merges (🟡 Minor) — `continue` instead of fall-through, plus a positive test asserting `__tombstone` is absent from surviving entry.
5. **Add an end-to-end resolver test** for `enforce_origin_tier` wire-in (🟡 Minor) — covers regression risk on future refactors.
6. **Tighten the docblock at `resolver.php:121-141`** — clarify customizable-filtering scope.

dabowman:
## Fixup pass for review feedback

Addressed all three 🟠 majors from the review. Pushed four commits on top of the existing branch.

### `filter_v3_block` list-shape corruption (Major #1)

Fixed in `87b47fb`. Added a companion walker `collect_list_shapes()` that records every path in the downstream block holding a keyed list (id/slug/name) or plain integer-indexed list. After the dot_set rehydrate, `restore_list_shapes()` walks those paths longest-first and converts the rehydrated assoc maps back into lists — re-embedding the captured key field on each entry for keyed lists, or `array_values()`-stripping for integer-indexed lists.

The `LIST_INDEX_SENTINEL` constant distinguishes integer-indexed lists from keyed lists in the shape map. New `dot_ref` / `dot_set_ref` helpers operate on segment-arrays (instead of dotted strings) so the empty-path case — when the block itself IS the list, e.g. `commands: [...]` at top level — round-trips through `container_path=null`.

Survival-path test covering the reviewer's exact concern: `commands.save.shortcut` allowlisted survives + downstream-merge over a 3-entry base preserves all 3 entries (the pre-fix list-replacement bug). Plus nested keyed-list `screens[id].apps[]` shape preservation.

### `src/apps/user-menu/index.js:27` user-switchable (Major #2)

Fixed in `8490250`. Flipped to the recommended canonical: PHP bridge in `wp_admin_shell_get_available_shells()` emits the kebab key `user-switchable` directly; JS reads `s['user-switchable'] || s.userSwitchable` (kebab wins, camelCase fallback for any plugin-supplied shell that hasn't migrated). Read-both fallback survives at READ time only — drops at v3.1 with the rest of the camelCase shims. `app.json` + `app.md` docs updated to match.

### `is_safe_href` whitespace + backslash variants (Major #3)

Fixed in `3fd722e`. `trim($href)` runs before any pattern check, so leading whitespace (HTML5-strippable space/tab/newline) no longer bypasses the `strpos('//') === 0` check. Defensive rejects added for `\\evil.example.com` (double backslash) and `\/\/evil.example.com` (escaped slash). Plus the previously-missing `vbscript:` reject assertion lock-in.

### What I did NOT address (out of scope per the review)

- `__tombstone: true` shape pollution (🟡 Minor) — flagged but not in the user's top-3 directive.
- Test coverage gap on REST `permission_check` with no permissions block (🟡 Minor).
- `enforce_origin_tier` integration coverage through `resolve_with()` (🟡 Minor).
- `shrink_against_baseline` / `enforce_trust_tiers` audit-emit dedup (🟡 Minor).
- `array('capabilities'=>[],'roles'=>[])` factory helper (🟡 Minor).
- All ⚪ Nits.

### Test results

All 19 PHP runners + `npm run test:schema` + `test:parity` + `test:runtime` + `test:engines` + `lint:js` clean.

`run-security-cascade-tests.php` expanded from 35 → 55 assertions (+20):
- 11 new list-shape preservation tests
- 9 new is_safe_href variant tests (whitespace + backslash + vbscript)

Test surface bumped to ~1645 in CLAUDE.md.

### Commits

- `3fd722e` fix(v3): is_safe_href trim() + reject backslash variants
- `8490250` fix(v3): kebab user-switchable across PHP→JS bridge + JS reader
- `87b47fb` fix(v3): preserve list-of-keyed-objects shape in filter_v3_block
- `9b09483` docs(CLAUDE): bump test-surface to ~1645

dabowman:
## Re-review — PR #61 fixup pass

### Overview

The fixup pass cleanly resolves all three 🟠 Majors from the original review with proportionate test coverage. The list-shape walker is the load-bearing fix and it's implemented carefully — longest-first restoration, sentinel for integer-indexed vs id-keyed shapes, segment-array helpers for the empty-prefix case where the block itself IS the list. I traced the edge cases (nested keyed lists, list-of-tuples in preload, mixed shapes, deny-list interaction with rehydrate, longest-first ordering invariant) and the algorithm holds. The kebab `user-switchable` rename is now consistent across all PHP/JS surfaces with a documented one-cycle camelCase fallback at READ time only. `is_safe_href` whitespace + backslash + vbscript coverage is solid. Test surface bumped 35 → 55 with twelve new attack-surface assertions, eight defense-in-depth assertions. **Mergeable** — one narrow Minor about FF (`\x0c`) trim coverage and the previously-deferred `__tombstone` pollution remain on the books for follow-up, but neither blocks v3.0.0-beta.1.

### Fixes verified

1. **🟠 List-shape corruption in `filter_v3_block`** — fixed at `class-wp-admin-shell-customizable.php:241-242` + 271-277. The `collect_list_shapes()` walker (line 294) records every list path the downstream block carried, distinguishing keyed lists (`id` / `slug` / `name`) from plain integer-indexed lists via `LIST_INDEX_SENTINEL`. `restore_list_shapes()` (line 350) walks captured paths longest-first so deeper relists complete before parents, re-embedding the key field on each entry. New `dot_ref` / `dot_set_ref` helpers operate on segment-arrays (not dotted strings) so the `container_path = null` case — when the block itself IS the list (`commands` at root) — round-trips correctly. End-to-end trace verified: `commands.save.shortcut` flows through filter → rehydrate → restore → merge against the 3-entry base list, all three base entries survive, `save` entry gets the new shortcut while retaining baseline `invoke`. `compileCommands.mjs:52` uses `Array.isArray(commands)` which would have silently skipped ALL commands on the assoc-map output — fix matters end-to-end.

2. **🟠 `user-switchable` JS callsite** — fixed across the full PHP→JS bridge:
   - `src/apps/user-menu/index.js:32` reads `s['user-switchable'] || s.userSwitchable` (kebab wins, camelCase fallback documented inline).
   - `wp-admin-shell.php:920` + `:930` emit kebab `user-switchable` directly.
   - `includes/class-wp-admin-shell-cli.php:46` + `:51` renamed CLI column to kebab; `app.json` / `app.md` doc strings updated.
   - Grep for `userSwitchable` across the PR branch shows only fallback-read sites + a v2-migration helper (intentional). All five PHP entry points use the canonical kebab → fallback camelCase pattern.

3. **🟠 `is_safe_href` whitespace + backslash bypass** — fixed at `class-wp-admin-shell-menu-items.php:484-538`. `trim()` runs before any pattern check (catches space, tab, LF, VT, CR, null byte). Explicit rejects added for `\\` and `\/\/` prefixes. `vbscript:` confirmed reachable via the `strpos(':') === false` fall-through and now asserted. Test coverage: leading-space, leading-tab, leading-newline, multi-whitespace mixed, `\\evil.example.com`, `\/\/evil.example.com`, `vbscript:`, plus a whitespace-padded-but-safe positive case.

### 🔴 Blockers

None.

### 🟠 Major

None remaining.

### 🟡 Minor

- **`is_safe_href` doesn't catch FF (`\x0c`).** PHP's `trim()` default character set (` \t\n\r\0\v`) omits form-feed. Per WHATWG, browsers strip FF from URLs before navigation. Verified: `php -r 'var_dump(trim("\x0c//evil.com"));'` → still has FF prefix, post-trim `strpos('//') === 0` is false. Narrow attack — FF in URLs is exotic — but the docblock claims HTML5 stripping equivalence. Fix: pass an explicit charlist `trim($href, " \t\n\r\0\x0b\x0c")` (adds FF; VT already covered by `\v`). Could lock in one more assertion to match.

- **`__tombstone: true` pollution on consumer-origin merges still latent.** `class-wp-admin-shell-merge.php:285` — after the `_doing_it_wrong` notice for a consumer-origin tombstone, execution falls through into `merge_internal( $entry, $over_index[$id], ... )` which deep-merges `__tombstone: true` from `$over_index[$id]` into the surviving base entry. Under `WP_DEBUG=false` this happens silently. Reviewer flagged in original review (🟡 Minor); fixup follow-up correctly noted as deferred. Defensible, but the fix is one line (`unset( $over_index[$id]['__tombstone'] );` before the merge call). Worth folding into Stage 5 hygiene.

- **`docs/upgrade-v2-to-v3.md` doesn't list the `userSwitchable` shim** alongside the other v3.0 → v3.1 deprecations. The JS code comment says "drops at v3.1" but the official deprecation timeline at `docs/upgrade-v2-to-v3.md:224` (the v3.0 / v3.1 column table) doesn't include it. Add one row for completeness; otherwise the v3.1 release-engineering grep won't catch the removal.

- **Test coverage gap: `restore_list_shapes` non-contiguous integer keys.** The walker handles them correctly (verified by trace — `[0=>x, 2=>y]` is `is_assoc=true`, gets `array_values()`'d to `[x, y]`), but no test exercises the case. Would happen if the allowlist denies entry [1] of an integer-indexed list, leaving a gap. Theoretical until a real `preload[]` allowlist scenario crops up.

### ⚪ Nits

- `dot_ref` line 400 has an extra space before the comma: `array_key_exists( $seg , $cur )`. Cosmetic.
- The `LIST_INDEX_SENTINEL` constant declaration sits between `restore_list_shapes()` and `dot_ref()`; conventional placement would be alongside `SENTINEL` near the top of the class.
- The list-shape capture walker re-traverses the downstream tree after `collect_leaves` already walked it — minor redundancy (two passes instead of one). Performance impact is sub-millisecond for realistic doc sizes; not worth restructuring.
- `merge_with_tombstones` migration in `run-cascade-tests.php` swapped four `merge()` calls for `merge_with_tombstones()` — semantics preserved (the bare `merge()` no-ops tombstones now per the trust-tier rules), so the contract being tested is unchanged, just routed through the right entry point. No assertion weakening.

### Out-of-scope deferrals

All six items the fixup follow-up noted as deferred were graded 🟡 Minor in the original review; none touch the security thesis or the v3 cascade gates.

- **`__tombstone: true` pollution** — defensible defer; one-line fix; surface area touched by this PR but the pollution is unrelated to list-shape correctness. Promote to Stage 5 hygiene PR.
- **REST `permission_check` no-permissions-block fixture** — defensible; covered indirectly by `run-cap-gating-smoke.php`'s default-permissions path; explicit assertion is cleanup, not coverage gap.
- **`enforce_origin_tier` integration coverage through `resolve_with()`** — defensible defer; the wire-in line `resolver.php:122` is one line and refactor risk is the legit concern. Worth one end-to-end fixture in Stage 5 to inoculate against drift.
- **`shrink_against_baseline` / `enforce_trust_tiers` audit-emit dedup** — defensible; pure maintenance debt, no behavioral concern.
- **`array( 'capabilities' => array(), 'roles' => array() )` factory helper** — defensible; trivial DRY win that doesn't affect correctness.
- **All ⚪ Nits** — agreed, defer.

### Regression hunting

- Walker performance: `collect_list_shapes()` recurses the same tree `collect_leaves()` walks — two passes O(N+N) instead of one. For realistic v3 docs (< 10 KB JSON), sub-millisecond. Not a concern.
- No test additions weaken prior assertions; cascade-tests delta is migration from `merge()` → `merge_with_tombstones()` for tombstone-bearing payloads (correct per the trust-tier model).
- `app.json` / `app.md` updates are docs-only.
- Webpack build / lint not exercised here, but the JS change is one-line + comment.
- No new lint waivers, no new `__experimental*` imports, no kernel DS-neutrality regression.

### Test coverage on the original review's specific gap

The original review flagged: "Trust-tier shrink-only rule — does the test catch DOWNGRADE of OR-semantic to AND-semantic via key removal?" On closer reading this isn't actually an attack vector — removing keys ALWAYS shrinks the OR-set (more restrictive, never escalates). The shrink-only invariant covers it by construction. No additional test needed.

The new 20 assertions break down as: 11 list-shape preservation (matching the reviewer's exact concern about `commands.save.shortcut` survival + nested keyed-list `screens[id].apps[]`) + 9 `is_safe_href` variants (3 whitespace + 2 backslash + vbscript + 3 positive-cases including whitespace-padded-safe + whitespace-only). All exercise actual attack surfaces, none are happy-path padding.

### Verdict

**Approve — minor follow-ups defensible to land separately.** Three 🟠 Majors closed cleanly with attack-surface coverage. Three 🟡 Minors (FF char, `__tombstone` pollution, upgrade-doc shim listing) carry forward to a hygiene PR without blocking v3.0.0-beta.1.

PR #62, opened 2026-05-22
v3: PR #49 review — runtime perf + schema/filter hygiene (Stages 3+5)

dabowman:
## Summary

Stages 3 + 5 of the [5-PR plan](docs/plans/2026-05-22-pr49-pre-merge-feedback.md) landing PR #49 pre-merge review feedback. Bundles JS runtime perf (2 items) + schema/filter hygiene (3 items).

### Stage 3 — JS runtime perf

- **`BindingsConsumer` rebinds keydown every render.** `src/runtime/bindings/BindingsConsumer.js` — `commands` array now built inside `useMemo([config?.commands, config?.bindings])`. Extracted `buildCommandsArray()` as module-scope helper. Document `keydown` handler binds once per real shortcut change instead of every router event.
- **`useDataView` cache unbounded.** `src/runtime/dataView/useDataView.js` — `cache` Map + `inflight` Map both routed through `lruSet()` with `LRU_CAP = 64`. Extracted `src/runtime/dataView/lruCache.mjs` for testability. Insertion-order eviction (Map preserves insertion order; oldest evicted first). Comment: cap sized for ~2× working set; tune after telemetry.

### Stage 5 — Hygiene

- **Schema `$id` numbering off-by-one.** v1 schema declared `$id: .../admin/v0.json`; v2 declared `.../admin/v1.json`. Renamed to match file versions. Added `$comment` noting prior identifier on each. Same fix applied to `admin-app-v2.json`, `admin-engine-v2.json`. `admin-v3.json` already correct.
- **Filter priority collision.** `wp_admin_shell_data` had `bind_screens()` + `inject_app_baselines()` both at priority 5 — execution order undefined. Bumped `inject_app_baselines()` to **6**. Reason: screens must exist before baselines attach. Documented in `CLAUDE.md` "Recurring patterns" + new "Core callback priorities" subsection in `docs/upgrade-v2-to-v3.md`.
- **Filter-vs-compiler ordering docblock.** Item 13 — skipped here. Stage 1 (sibling PR `feat/v3-security-cascade`) owns this. Verified via `git show`.

## Locked decisions honored

- `useDataView` LRU cap: **64 with code comment "tune after telemetry"**
- Compiler-vs-filter ordering: **document current order (option B)** — filter receives author shape, compile runs after. Plugin route synthesis uses per-origin filters at priority 5.

## Test surface delta

**New runtime tests (+22 assertions):**
- `tests/runtime/bindings-consumer-rebind.test.mjs` (13) — pure-function tests on `buildCommandsArray()` + static-analysis guards on source (`useMemo` deps shape + extracted helper).
- `tests/runtime/data-view-lru.test.mjs` (9) — overflow eviction, update-doesn't-promote, chainable, `CAP=1` edge.

**Merge gate (all green):**
- `npm run test:schema` — 103 (unchanged; schema `\$id` rename had zero blast radius since validator uses file path, not `$id`)
- `npm run test:runtime` — 532 (was 510, +22)
- `npm run test:parity` — 4
- `npm run test:engines` — 77
- `npm run lint:js` + `lint:ts` clean
- `npm run build` clean

PHP test sweep not run inside the worktree (port-bound wp-env). Filter-priority change is mechanical — `inject_app_baselines()` invoked directly in `run-data-view-tests.php` so registration priority is invisible. PHP-side smoke recommended pre-merge.

## Test plan

- [ ] `npm run test:runtime` — confirm 532 / 0 fail
- [ ] PHP sweep — `run-cascade-tests.php`, `run-data-view-tests.php`, `run-classic-menu-bridge-tests.php`, `run-menu-route-shims-tests.php` all green
- [ ] Manual: open every entity-CRUD app + switch shells — no console errors, no visible re-bind thrash
- [ ] DevTools event-listener panel: confirm single `keydown` listener after router navigation
- [ ] Plugin authoring smoke: contribute a screen via `wp_admin_shell_data_plugin` callback — confirm `inject_app_baselines` attaches dataView baseline (validates priority 5 → 6 ordering)
- [ ] External Ajv-using consumer: confirm v1/v2 `$id` rename doesn't break (none known; `$comment` left for migration)

## Plan deviations

- `bindings-consumer-rebind.test.mjs` design: plan suggested JSDOM-shaped React test stubbing `document.addEventListener`. No JSDOM scaffold in repo. Took the "pragmatic over perfect" path the plan flagged — pure-helper tests + static-analysis guards on source. Equivalent coverage; regressions still caught via the `useMemo` deps shape assertion.
- `tests/schema/validate-shells.test.mjs` uses `compileSchema()` per file path, not via `\$id` resolution. So `\$id` rename had zero blast radius. 103 schema assertions unchanged.


dabowman:
## Code Review — PR #62 runtime perf + schema/filter hygiene

### Overview

Tight, well-scoped bundle. Five PR-49 review items shipped: `BindingsConsumer` `useMemo` rebind fix, `useDataView` LRU cap at 64, schema `$id` off-by-one reconciliation (v1/v2 + v2 manifests), `wp_admin_shell_data` priority pin (5 → 6 for `inject_app_baselines`), and the CLAUDE.md / `docs/upgrade-v2-to-v3.md` contract docs. Plan-doc adherence is high — both locked decisions honored (LRU cap of 64 with telemetry note; document-current-order on compiler-vs-filter). +22 runtime assertions across two new test files. The lone test-design deviation (static-analysis + pure-function instead of JSDOM mount for `BindingsConsumer`) is defensible and well-flagged.

### 🔴 Blockers

None. The PR is mergeable on its own merits.

### 🟠 Major

- **No actual coverage that `useMemo` removal causes a regression.** `tests/runtime/bindings-consumer-rebind.test.mjs:60-67` runs four regex assertions on the source string. They prove the deps array literal `[ config?.commands, config?.bindings ]` and `}, [ compiled ]` exist — but a regression where someone removes the `useMemo` wholesale and inlines `compiled` back into the effect would *also* fail these regexes, which is fine. The real concern: a future refactor that legitimately renames `compiled` to e.g. `entries` would also fail. So the test guards the *shape* not the *behavior*. That's an acceptable trade-off — but please own the limitation in the test header comment. The current header says "we accept the slight duplication" referring to the helper extraction, not the static-analysis brittleness. One sentence under "Static-analysis guards" acknowledging "these will need updating on any var-name refactor" makes the trade-off explicit for the next contributor.

### 🟡 Minor

- **`buildCommandsArray` not exported from a `.mjs` file.** Per CLAUDE.md "Pure-JS runtime modules go in `.mjs` files." `buildCommandsArray` is pure / side-effect-free per its own JSDoc, but it lives in `BindingsConsumer.js` because the React component needs to consume it from the same module. The test then has to reconstruct the function via `new Function( source-extraction )` (`bindings-consumer-rebind.test.mjs:73-82`) — the test header itself flags "cleaner alternative: split `buildCommandsArray` into its own .mjs file." That cleaner alternative matches existing repo convention (`resolveRegion.mjs`, `matchRoute.mjs`, `lruCache.mjs` in *this same PR*). Suggest: extract to `src/runtime/bindings/buildCommandsArray.mjs`, import from both `BindingsConsumer.js` and the test, drop the Function-constructor extraction. The test would shrink + lose its brittleness on whitespace inside the regex match.

- **Inflight LRU eviction can drop a still-in-flight fetch from the dedup map.** `useDataView.js:178` (`lruSet( inflight, key, promise, LRU_CAP )`). The closure-captured `promise` keeps the fetch alive and the consumer's `.then` still runs, but a concurrent caller for the same `key` that runs between eviction-from-inflight and the original promise settling will not find the entry via `inflight.get( key )` and will fire a duplicate fetch. With `LRU_CAP = 64` and inflight Map sized only on actually-pending fetches, that requires >64 concurrent in-flight fetches — pathological in practice. Mention it as a tracked-but-accepted edge case in the code comment, or hold off bounding `inflight` at all (it's already self-limiting on resolution via the `.finally` delete). Bounding `cache` is the load-bearing change; bounding `inflight` is belt-and-suspenders that adds a (tiny) failure mode.

- **`LRU_CAP = 64` hardcoded with no override.** The comment "tune after telemetry" implies telemetry will arrive — but there's no mechanism for a site author or test to override the cap without forking the module. A constant export `export const LRU_CAP = 64;` would let `tests/runtime/data-view-lru.test.mjs` import the canonical value instead of redeclaring `const CAP = 64` (`data-view-lru.test.mjs:39`). That drift means if someone bumps `LRU_CAP` to 128 in `useDataView.js`, the LRU test still asserts against 64 and silently passes. Cheap fix: export the constant from `lruCache.mjs` (or `useDataView.js`) and import it in the test.

- **`wp_admin_shell_data` priority table covers 2 of 3 actual callbacks.** The docs (`docs/upgrade-v2-to-v3.md:223-232` + CLAUDE.md "Recurring patterns") list priorities 5 / 6 / 999 — but they elide that priority 999 is `warn_legacy_view_configs`, which already exists at `class-wp-admin-shell-data-view-config.php:881`. Both the CLAUDE.md addition and the docs table list it correctly, so this is consistent. ✅ Looks good on reread — initially misread the diff. (Leaving the note in place since I verified.)

- **`bindings-consumer-rebind.test.mjs` uses ad-hoc `ok()` instead of `node:assert/strict`.** The PR-49 review explicitly flagged `tests/runtime/mode-resolution.test.mjs:29-42` for the same thing. This new test perpetuates the pattern. Not a blocker — every other runtime test in the suite uses ad-hoc helpers — but a missed opportunity to set the trend in the right direction for new tests.

### ⚪ Nits

- **Test header comment in `bindings-consumer-rebind.test.mjs:11`** says "The runtime tests are pure-ESM-leaning by convention." Stronger: they are pure-ESM by convention, full stop. The test itself proves the rule by working around it (`new Function(...)` extraction).

- **`lruCache.mjs:22` "NOT access-order" comment** could note that this is intentional, not a TODO. Suggest: "By design — re-reading an entry doesn't promote it. dataView triples are read-heavy and short-lived, working-set churn is on shell-switching, not within-session re-access." (Echoes the design rationale already in the file's top comment — restating at the assertion site removes confusion for a drive-by reader.)

- **`buildCommandsArray` JSDoc** documents that `null`/`undefined`/non-array inputs return `[]`, but the actual code at line 19 also returns `[]` for `[]` (empty array). Spelling that out in the JSDoc — "empty / nullish / non-array inputs" — closes a tiny gap.

- **CLAUDE.md addition at line 147** uses the phrase "core callback priorities are contractual" — strong phrasing. Maybe "documented" or "fixed" is closer to the actual commitment. "Contractual" implies SemVer-breaking on change; the docs themselves say "do not assume any other interleaving" which is weaker. Pick one tone.

- **Test surface delta in CLAUDE.md** updated from "~1580 assertions" → "~1602 assertions, post-3d.3 / PR #60 feedback + PR-49 perf/hygiene bundle" — but the canonical phrasing across CLAUDE.md is roundabout. Suggest just "~1602" without re-stating the provenance every status block; the git log is the provenance.

**Defensible.** The plan (stage 3) flagged JSDOM as the ideal-but-absent path: "plan suggested JSDOM-shaped React test stubbing `document.addEventListener`. No JSDOM scaffold in repo." The PR author chose the pragmatic alternative — pure-function tests on the extracted helper + static-analysis guards on the source — and explicitly documented the trade-off in the PR body.

What's actually undertested as a result:

- The `useEffect` cleanup function (`document.removeEventListener( 'keydown', onKey )` at line 102) — if someone removes the cleanup return, the static-analysis grep wouldn't catch it.
- The `[ compiled ]` deps array's reactivity — if React's `useMemo` semantics change (extremely unlikely) or if `useMemo` is replaced with a different memoization helper that has different semantics, no test catches it.
- The interaction between `compiled` referential stability and the `useEffect` dep array — the test asserts the literal text `}, [ compiled ]` exists; it does not assert that `useMemo` returns a stable reference across renders when deps don't change.

All three gaps are unlikely-to-regress in the current shape. Reasonable to defer until a JSDOM scaffold lands (issue #30 tracks this). Make sure that issue references this PR's test-pattern as a deferred coverage gap.

### What's right

- **Filter-priority documentation propagated to three sources** (`CLAUDE.md` "Recurring patterns" → `docs/upgrade-v2-to-v3.md` priority table → docblock comments on both `bind_screens` and `inject_app_baselines` registrations). All three sources agree. That's the kind of contract documentation that survives a year of churn.

- **Schema `$id` rename is symmetric and complete.** Verified all five affected schemas (`admin-v1`, `admin-v2`, `admin-app-v2`, `admin-engine-v2` — the rename targets — plus `admin-v3` / `admin-app-v3` / `admin-engine-v3` / `tokens-v1` which were already correct). No cross-schema `$ref` references to old URLs (all internal `#/$defs/...`). The `$comment` field at each renamed `$id` correctly records the prior identifier for forward-compat.

- **LRU update-doesn't-promote semantics are correct + tested.** `lruSet`'s key check `! map.has( key ) && map.size >= cap` is exactly right — `Map.set` on existing key preserves insertion order, so the update path never reaches the eviction branch. The test at `data-view-lru.test.mjs:65-82` covers the subtle case explicitly.

- **Module-scope `buildCommandsArray` extraction** is genuinely cleaner than the prior inline-in-`useEffect` shape — the helper is now reusable, testable, and the React component reads more linearly. Even setting aside the perf fix, this is a code-clarity win.

- **`inflight` LRU correctly handles the "key already present" path** — concurrent callers for the same key still share the same in-flight promise. The `Map.set` semantics are load-bearing here and the comment in `lruCache.mjs:18` explicitly calls this out.

- **`buildCommandsArray` empty-input behavior is correct** — returns `[]` (referentially-stable empty array would be even better, but `useEffect`'s `compiled.length === 0` short-circuit at line 60 makes the ref churn moot in practice).

- **All four PR-49 review items addressed in this stage actually addressed** — checked against the original review thread. Item 6 (BindingsConsumer rebind) ✅. Item 7 (useDataView unbounded cache) ✅. Item 9 (schema `$id` collision) ✅. Item 10 (filter priority collision) ✅. Item 13 deferred to sibling PR per body — confirmed no duplication.

1. **Extract `buildCommandsArray` to `src/runtime/bindings/buildCommandsArray.mjs`** — matches `.mjs` convention, eliminates the `new Function( source-slice )` test extraction, makes the test less brittle to source-text changes. ~3-line change.
2. **Export `LRU_CAP` constant** so the LRU test imports it instead of redeclaring. Catches future drift.
3. **One-sentence honesty note in `bindings-consumer-rebind.test.mjs` header** about the static-analysis-guards-not-behavior-guards trade-off.

Everything else is style / nits — ship it.

dabowman:
## Review follow-up — 3 fixes + 1 nit

Addressed the three top-priority items from the code review:

### 1. `inflight` LRU bound removed (b8efdc6)

Reviewer's concern about the dropped-fetch edge case at >64 concurrent fetches. The `inflight` Map is now plain `Map.set()` — self-limiting via the `.finally( inflight.delete( key ) )` cleanup that was already in place. The persistent `cache` Map keeps its LRU cap (the load-bearing change). Code comment on the Map declaration documents the rationale.

### 2. `LRU_CAP` exported as a constant (b8efdc6)

`export const LRU_CAP = 64` now lives in `lruCache.mjs` alongside `lruSet`, and is re-exported from `useDataView.js`. The LRU test imports the canonical value instead of redeclaring `const CAP = 64`. Added a sanity assertion that pins the value at 64 — an unintentional bump shows up in the test diff.

### 3. `buildCommandsArray` extracted to sibling `.mjs` (c60cdb2)

New `src/runtime/bindings/buildCommandsArray.mjs` owns the helper. `BindingsConsumer.js` imports + re-exports it for source-surface stability. The rebind test now imports the helper directly — gone is the `new Function( source-slice )` extraction hack and its brittleness on whitespace inside the regex.

Test header comment rewritten to honestly own the static-analysis trade-off (regression tripwire for *the specific shape* of the fix, not its *behavior* — a future legitimate `compiled` → `entries` rename would trip it). Also lists the three coverage gaps that need JSDOM mount tests, deferred to issue #30.

### 4. CLAUDE.md nit (a013873)

"core callback priorities are contractual" → "documented". Matches the softer tone the upgrade doc itself uses ("do not assume any other interleaving").

## Test counts after this batch

| Suite | Before | After |
|---|---|---|
| `test:runtime` | 533 | 532 |
| `test:schema` | 103 | 103 |
| `test:parity` | 4 | 4 |
| `test:engines` | 77 | 77 |

Net runtime delta: −1. `bindings-consumer-rebind` lost the `helper extracts from source` assertion (no longer needed — helper imports directly); `data-view-lru` gained a `LRU_CAP === 64` sanity assertion. The CLAUDE.md test-surface block already reads 1602 so no number update needed there.

Lint clean: `npm run lint:js` + `npm run lint:ts`.

## Deliberately not touched

- JSDOM scaffold (reviewer flagged as out-of-scope; followup tracked in issue #30 via the test header comment).
- Schema `$id`s and filter priority numbers (already correct).
- `node:assert/strict` migration of the test (reviewer flagged as missed opportunity not blocker; would also touch `mode-resolution.test.mjs` and break the in-flight pattern across the suite — better as a dedicated cleanup PR).
- `docs/upgrade-v2-to-v3.md` tone (out of the agreed change scope).

dabowman:
## Re-review — PR #62 fixup pass

### Overview

Three commits — `b8efdc6` (drop `inflight` LRU bound + export `LRU_CAP`), `c60cdb2` (extract `buildCommandsArray` to sibling `.mjs`), `a013873` (CLAUDE.md tone nit) — cleanly address the top three priorities from the original review plus the one ⚪ nit. Verified all suites green (runtime 532, schema 103, parity 4, engines 7, lint clean). Fixup is tight; no regressions surfaced.

### Fixes verified

- **`inflight` LRU bound dropped (b8efdc6).** `useDataView.js:192` is plain `inflight.set( key, promise )`. The pre-existing `.finally( inflight.delete( key ) )` at line 185–187 keeps the map self-draining. Inline comment at the `set()` site + the Map declaration both document the rationale. The pathological dedup-miss edge case the original review flagged is now structurally impossible — no `lruSet` call on `inflight` anywhere.
- **`LRU_CAP` exported (b8efdc6).** `export const LRU_CAP = 64` lives at `lruCache.mjs:11`, re-exported from `useDataView.js:13`. Test imports the canonical value via `LRU_CAP: CAP` rename-on-import. Sanity assertion `ok( 'imported LRU_CAP equals 64 (default sizing)', CAP === 64 )` at `data-view-lru.test.mjs:40` — silent drift on future bumps now impossible.
- **`buildCommandsArray` extracted (c60cdb2).** New `src/runtime/bindings/buildCommandsArray.mjs` (32 lines, single export, pure ESM, imports only `parseShortcut` sibling). `BindingsConsumer.js` imports + re-exports the helper (re-export keeps the public surface stable). Test imports the helper directly via `await import( … )` — the `new Function( source-slice )` extraction hack is gone. Signature + edge-case handling preserved (verified diff of helper body: lines 20–32 of new file match the prior inline lines 17–28 byte-for-byte modulo the `export` placement). JSDoc now also explicitly lists empty arrays as a returns-`[]` case ("empty, nullish, or non-array inputs" — closes the prior ⚪ nit).
- **Test header honesty (c60cdb2).** Header at `bindings-consumer-rebind.test.mjs:2–39` now has the requested machine-readable structure: two-layer description (functional behavior vs static-analysis tripwires), explicit "These are regression tripwires for *the specific shape* of the fix, not behavioral guarantees" callout, three named coverage gaps (cleanup-function firing, `useMemo` referential stability, `compiled`/dep-array interaction), and a direct reference to issue #30 for the JSDOM-scaffold tracking. Future contributor searching for `JSDOM`, `issue #30`, or `coverage gaps` lands on the right block.
- **CLAUDE.md `contractual` → `documented` (a013873).** Single-line `s/contractual/documented/` at line 146. No other content drift in the file. Tone now matches the upgrade doc's "do not assume any other interleaving" softer phrasing.

### Test coverage validation

| Suite | Claimed | Actual |
|---|---|---|
| `test:runtime` PASS count | 532 | 532 ✅ |
| `test:schema` | 103 | 103 ✅ |
| `test:parity` | 4 | 4 ✅ |
| `test:engines` | 77 | 77 ✅ |
| `bindings-consumer-rebind.test.mjs` | 12 (was 13, −1 from helper-extracts-from-source removal) | 12 ✅ |
| `data-view-lru.test.mjs` | 10 (was 9, +1 from LRU_CAP sanity) | 10 ✅ |

Net runtime delta `+22 → +21` matches the fixup follow-up's accounting (`bindings −1`, `LRU +1`). CLAUDE.md test-surface block lists both new tests by name at line 188 (`bindings-consumer-rebind` + `data-view-lru`).

### Regression hunting

Nothing surfaced:
- Helper body bit-identical pre/post extraction — no subtle behavior drift (no closure over module state, no React import in the `.mjs`, no side effects).
- No console.log / debugger / new lint suppressions in any of the fixup-touched files.
- No coverage drop on original semantics: the `helper extracts from source` assertion (removed) was scaffolding for the extraction hack itself, not a behavioral guarantee — its removal alongside the hack is symmetric and intended.
- PR description / body claims match the actual commits (verified against `git log` + `git show --stat`).
- The four static-analysis tripwires on `BindingsConsumer.js` still hold against the post-extraction source (verified: the `export { buildCommandsArray };` re-export at line 12 matches the new tripwire regex `/export \{ buildCommandsArray \};/`).

### 🔴 Blockers

None.

### 🟠 Major

None.

### 🟡 Minor

None.

### ⚪ Nits

- ⚪ Pure aesthetics: `lruCache.mjs` now has two top-level JSDoc blocks (one on `LRU_CAP`, one on `lruSet`). Both are well-written; together they slightly bury the actual function declaration. A drive-by reader might appreciate a one-line `// ---` separator. Take it or leave it.
- ⚪ The new `inflight` Map docblock (`useDataView.js:23–31`) is excellent — 9 lines of "why no LRU" rationale at the declaration site. Mention of the >64 threshold as "pathological in practice" addresses the prior review's edge-case ask honestly without over-engineering it.

### Verdict

**Approve — ship it.** Three commits, all clean. No remaining blockers, majors, or minors. The pre-existing 🟡 minors that weren't part of this fixup scope (`node:assert/strict` migration, `lruCache.mjs:22` design-note nit) were already addressed in this pass or explicitly deferred by the author with a defensible rationale. Test-surface accounting is honest and matches reality.

PR #63, opened 2026-05-22
v3: PR #49 review — kernel DS-neutrality cleanup (Stage 4)

dabowman:
## Summary

Stage 4 of the [5-PR plan](docs/plans/2026-05-22-pr49-pre-merge-feedback.md) landing PR #49 pre-merge review feedback. Relocates `WpdsThemeProvider` + `SlotFillProvider` out of kernel into engine code; the kernel becomes truly DS-neutral, validated by a static-analysis test.

## Items addressed (from PR #49 review #3)

- **🟠 B1 — Kernel imports `WpdsThemeProvider`.** Previously `src/runtime/styles/ThemeProviderHost.js:28` imported the WPDS-flavored provider as both the absent-engine fallback and the error-boundary recovery target. Relocated to `src/runtime/engines/core-default/WpdsThemeProvider.js`. `core:default` ships it via `EngineSource.ThemeProvider`. `core:single-pane` reuses it via sibling import; documented dependency in its `index.js`. `core:desktop` opts out (no `ThemeProvider` declared) — renders through the new `NeutralProvider` pass-through wrapper, which mounts no DS provider.
- **🟠 B2 — Kernel imports `SlotFillProvider`.** `src/runtime/kernel.js:7` previously pulled `@wordpress/components`. Relocated the `<SlotFillProvider>` wrap into each engine's `Layout.js` (`core-default`, `core-single-pane`, `core-desktop`). Kernel is fully DS-neutral.
- **⚪ N1 — DOM attribute rename.** `data-wpds-theme-provider-id` → `data-theme-scope-id` everywhere (host, engine compileStyles, source-types JSDoc). Sibling `data-wpds-shell-detail` → `data-theme-scope-detail`. Drops misleading DS-specific prefix; engine theming layer no longer leaks WPDS naming into kernel-emitted DOM.

## Architectural commitments

- Engines opt into shell theming by shipping a working `EngineSource.ThemeProvider`. **Absence renders un-themed via neutral pass-through.** Kernel never silently injects WPDS (or any DS) fallback.
- Error-boundary recovery (when engine-supplied provider throws): swap to neutral pass-through, NOT WpdsThemeProvider. Engines opt into recovery by shipping a working provider.
- Slot/Fill substrate is engine-side. Bundled engines wrap `<SlotFillProvider>` in `Layout.js`; non-WPDS engines without `@wordpress/components` consumers can omit.

## Test surface delta

**New runtime tests (+240 assertions):**
- `tests/runtime/theme-provider-host.test.mjs` (40) — pure-helper coverage on extracted `themeScope.mjs`: `pickDensity` tier precedence, `hasThemeContent` recognition, `scopedSelector` prefix routing, `buildScopedDetailCss` emits renamed `data-theme-scope-id` selector across all three buckets.
- `tests/runtime/kernel-no-ds-import.test.mjs` (200) — **static-analysis guard** that codifies the CLAUDE.md kernel-DS-neutrality contract. Scans every kernel file (`kernel.js`, `index.js`, `ThemeProviderHost.js`, `Region.js`, plus support modules) for forbidden imports of `@wordpress/components`, `@wordpress/ui`, `@wordpress/icons`, `@wordpress/dataviews`, and the relocated `WpdsThemeProvider` paths. Future regressions surface immediately.

**Merge gate (all green):**
- `npm run test:schema` — 103
- `npm run test:runtime` — 750 (was 510; +240)
- `npm run test:parity` — 4
- `npm run test:engines` — 77
- `npm run lint:js` + `lint:ts` clean
- `npm run build` clean (216 KiB entry + 22 KiB CSS, unchanged from baseline; old `data-wpds-theme-provider-id` selectors absent from bundled output — `node_modules/@wordpress/theme/.../design-tokens.css` still emits it for its own internal wrapper, distinct from kernel scope wrapper)

## Slot/Fill consumer audit

`grep -rn "createSlotFill\|<Slot \|<Fill \|useSlot" src/` — exactly one consumer: `src/apps/simple-editor/index.js:405` renders `<Slot name=\"core:editor.sidebar\" .../>`. Mounts inside a region inside engine Layout → gets relocated `SlotFillProvider` substrate. `@wordpress/components` overlay components (Modal, Popover, Notice, SnackbarList) all use Slot/Fill internally — also live inside engine Layouts via mounted apps.

`NavigationGuard` and `BindingsConsumer` (the two kernel-side components rendered before `<Engine>`) verified DS-neutral.

## Test plan

- [ ] `npm run test:runtime` — confirm 750 / 0 fail
- [ ] Build smoke + grep one chunk: `grep -c WpdsThemeProvider build/index.js` — expect 0 in kernel chunk
- [ ] Manual: each bundled shell renders w/ theming intact (default, single-pane, desktop)
- [ ] Manual: notices snackbar fires correctly (Slot/Fill substrate works)
- [ ] Manual: command palette opens (Modal Slot/Fill)
- [ ] Manual: toolbar actions render in correct slot
- [ ] Manual: simple-editor opens with sidebar Slot working
- [ ] Manual: trigger engine ThemeProvider error (synthetic dev test) — neutral wrapper renders, shell still paints
- [ ] `core:desktop` engine: confirm windows render w/o WPDS theming (NeutralProvider path). Windows still consume `--wpds-*` tokens via `@wordpress/ui`'s built-in defaults — visually no regression since desktop had no seeds anyway

## Plan deviations

- **Pure-helper extraction.** Extracted `themeScope.mjs` sibling to make tests viable without JSDOM. Behavior identical. `.mjs` pattern matches existing codebase convention (`resolveRegion.mjs`, `matchRoute.mjs`).
- **`core:desktop` behavior change.** Pre-PR, desktop silently rendered through `WpdsThemeProvider` fallback. Post-PR, renders through `NeutralProvider`. Visually no-op since desktop has no seeds — only loss is shell-level seed propagation into windows, which desktop never used. Documented in `core-desktop/index.js` docstring + `docs/engines-and-design-systems.md`.
- **No `data-wpds-theme-provider-id` compat shim.** Search confirmed no in-tree consumers outside the host itself. External code (if any) needs to migrate; no shim added.

## Known coverage gap

React-side `<ThemeProviderHost>` (engine `ThemeProvider` priority, neutral-wrapper fallback, error-boundary recovery) lacks a JSDOM-level mount test — no JSDOM scaffold in repo. Pure-helper test covers the contract surface; manual browser smoke covers the React side. Same gap tracked alongside `<PersistentRegion>` `data-app-mounted` assertion (also no JSDOM yet).


dabowman:
## Code Review — PR #63 kernel DS-neutrality cleanup

### Overview

Stage 4 of the PR #49 follow-up chain — relocates `WpdsThemeProvider` from `src/runtime/styles/` to `src/runtime/engines/core-default/`, drops `SlotFillProvider` from `src/runtime/kernel.js`, wraps `<SlotFillProvider>` inside each engine's `Layout.js`, renames `data-wpds-theme-provider-id` → `data-theme-scope-id` (plus the sibling `data-wpds-shell-detail` → `data-theme-scope-detail`), extracts five pure helpers into a new `src/runtime/styles/themeScope.mjs`, and lands two static-analysis runtime tests (40 + 200 assertions). Net +749 / −306 across 17 files.

Highest-blast-radius PR in the chain. I stress-tested four surfaces hardest: (a) completeness of the kernel-file allowlist in `kernel-no-ds-import.test.mjs`, (b) whether the `FORBIDDEN_PATTERNS` regex catches every import shape, (c) `core:desktop`'s behavior change from silent WPDS fallback to `NeutralProvider`, and (d) Slot/Fill substrate placement vs every consumer in the React tree. Build runs clean (220 KiB entry / 22 KiB CSS — within rounding of PR's claimed 216 KiB / 22 KiB); `WpdsThemeProvider`'s consent string + allowlist literal both land in the entry chunk (eager `core-default` engine pulls it in), and the old `data-wpds-theme-provider-id` literal is absent from emitted JS while `data-theme-scope-id` is present. All 21 runtime test files exit clean.

### Blockers

None. The refactor is faithful to the plan, the move is a pure `git mv` (zero content diff between old and new `WpdsThemeProvider.js`), the Slot/Fill audit holds up under independent verification, and the contract test catches what it claims to catch.

### Major

- **🟠 `kernel-no-ds-import.test.mjs` `KERNEL_FILES` allowlist is incomplete.** 24 files listed; cross-referencing `git ls-tree origin/feat/v3-kernel-ds-neutrality src/runtime/` (minus `engines/`) shows **13 kernel files unlisted**: `bindings/parseShortcut.mjs`, `bindings/triggerStore.mjs`, `dirty-state/dirtyState.mjs`, `dirty-state/useDirtyState.js`, `dashboardGrid/composeWidgets.mjs`, `dataView/{deprecation,hydrateInline,mergeFields,useDataView}.{mjs,js}`, `manifests/index.js`, `modes/{resolveMode.mjs,useMode.js}`, `registry/builtins.js`, `shell-switching.js`, `tokens/tokensResolver.mjs`. I checked each — none currently import a forbidden package, so the test is green by coincidence. But the kernel-DS-neutrality contract applies to all of them; a future drift in any (e.g. someone adding `import { Icon } from '@wordpress/icons'` to `useDataView.js` to render a "no data" placeholder) wouldn't be caught. The fix is mechanical: append the 13 paths to `KERNEL_FILES`. While there, consider switching to an auto-discovery pattern (`fs.readdirSync` walk of `src/runtime/` minus `engines/`) so new kernel files are picked up automatically — manual lists drift by definition.

- **🟠 `FORBIDDEN_PATTERNS` only catches `from "..."` syntax.** Every regex is anchored on `/from\s+['"].../`. This catches the canonical `import { X } from '...'` and `import X from '...'` shapes but **misses side-effect imports** (`import '@wordpress/components/build-style/style.css'`), CommonJS (`require('@wordpress/components')`), and dynamic imports (`import('@wordpress/components')`). The side-effect import is the highest-probability regression — someone adds an `import '@wordpress/components/build-module/notice/style.css'` line to a kernel file to silence a CSS warning and the contract test stays green. Suggested regex augmentation: add a sibling pattern matching `(?:import|require)\s*\(?\s*['"]@wordpress\/(?:components|ui|icons|dataviews)/` that picks up bare-side-effect imports, `require()`, and dynamic `import()`.

### Minor

- **🟡 `buildScopedDetailCss` hardcodes the attribute name as a string literal instead of consuming the exported constant.** `themeScope.mjs:134` writes `const scopeSel = ` `[data-theme-scope-id="${ providerId }"]`;`. The constant `THEME_SCOPE_ATTRIBUTE` is defined and exported just below (`themeScope.mjs:177`) but not consumed by the function that's supposed to be its source of truth. If the constant ever changes (e.g. the post-Stage-4 follow-up renames it again), the selector won't follow. The fix is one line: `` `[${ THEME_SCOPE_ATTRIBUTE }="${ providerId }"]` ``. Same applies to the historic-context line in the `ThemeProviderHost.js` docstring (line 15) — fine as-is since it's prose, but worth noting.

- **🟡 PR's test-count math is off.** PR body claims +240 runtime assertions for a total of 750. My run summed actual `PASS: N` lines across all 21 test files: **629**, plus the 40 + 200 from the two new tests = 869. Discrepancy is because 5 of the 26 test files in the script don't print a `PASS: N` line at the end (they use `OK / OK / OK` markers or similar). The CLAUDE.md update on line 15 of the diff says "750" — the right number depends on which counting convention you adopt, but it's not 750 by either count. Pick one (sum-of-`PASS:` or sum-of-`ok`-markers) and update CLAUDE.md to match before merge.

- **🟡 `core:desktop` chrome still imports `--wpds-font-family-sans`.** `src/runtime/engines/core-desktop/index.css:23` reads `font-family: var( --wpds-font-family-sans, -apple-system, system-ui, sans-serif );`. With `NeutralProvider` mounted there's no engine-level provider populating `--wpds-*` tokens, so the var resolves via `@wordpress/ui`'s passive CSS-layer defaults (which load globally via the apps' own `@wordpress/ui` imports). The robust 3-tier fallback (`-apple-system, system-ui, sans-serif`) means visually this is a no-op as the PR claims, but the implicit dependency on `@wordpress/ui`'s side-effect CSS load is fragile — a desktop-only build with zero WPDS-flavored apps would no longer set `--wpds-font-family-sans` from anywhere. Either drop the `--wpds-*` token-reference (since desktop opts out of WPDS) or document the implicit dep in the `engines/core-desktop/index.js` docstring. Low priority; not blocking.

- **🟡 No build-output assertion in CI.** PR's manual smoke checklist includes `grep -c WpdsThemeProvider build/index.js — expect 0 in kernel chunk` but no automated test enforces this. A trivial follow-up: add a `tests/build/no-wpds-in-entry.test.mjs` (or fold into existing build tests if present) that runs `npm run build` and greps for forbidden literals in the entry chunk after build. Defers the verification from human-discipline to CI. Not blocking — the static-analysis test already covers the source-level contract; this would just add belt-and-suspenders at the bundled-output level.

### Nits

- **⚪ Test-file headers list `THEME_SCOPE_DETAIL_ATTRIBUTE` as part of contract coverage but no test asserts it's consumed by `buildScopedDetailCss`.** The test asserts the constant equals `'data-theme-scope-detail'` (`tests/runtime/theme-provider-host.test.mjs:374`), but `buildScopedDetailCss` doesn't use it — that attribute is only emitted in `ThemeProviderHost.js`'s React render via `createElement( 'style', { [ THEME_SCOPE_DETAIL_ATTRIBUTE ]: id }, ... )`. The constant test alone is a weak signal — consider adding a string-match assertion that the test fixture invoking `buildScopedDetailCss` emits CSS scoped exclusively to `data-theme-scope-id` (already done) AND that no test path emits `data-theme-scope-detail` (a no-op check that catches accidental constant swap).

- **⚪ `ThemeProviderHost.js:15` docstring carries a historical-context line: "Pre-v3 the attribute was named `data-wpds-theme-provider-id`."** Fine for the diff cycle (helps reviewers); consider trimming once the PR lands and the rename is settled. Future code archaeology can find it in git history.

- **⚪ `engines/core-single-pane/index.js` reuses the sibling import pattern (`import { WpdsThemeProvider } from '../core-default/WpdsThemeProvider'`).** Documented in the file with a comment block explaining the tradeoff. This is fine for the bundled-engine case; if a third-party engine plugin authors a single-pane variant outside the shell repo, they'd need to either ship their own copy or import the path explicitly. Not a regression — it just makes `core-single-pane` formally depend on `core-default`'s WPDS provider being shipped. Captured in the doc; no change needed.

- **⚪ `EngineSource.ThemeProvider` JSDoc in `source-types.js:96` updated to `data-theme-scope-id={id}` — but the contract sentence still reads "Children must be rendered inside a wrapper carrying `data-theme-scope-id={id}` so shell-level scoped detail CSS can target them. (The kernel emits this wrapper around the provider; engines render their children as siblings of the kernel-supplied scoped wrapper.)"** Worth re-reading — the parenthetical says the kernel emits the wrapper, but the JSDoc opener says the engine's children "must be rendered inside" it. Both true today (the host wraps; the engine provider's children render inside that wrap), but the phrasing could be clearer.

1. **Pure-helper extraction into `themeScope.mjs`.** ✅ Defensible. The helpers are pure functions (no React, no DOM), behavior is identical to the inlined version (the diff is mechanical: copy out, re-import), and the pure module unblocks the 40-assertion test which would otherwise need a JSDOM scaffold the repo doesn't yet have. The `.mjs` convention matches existing pure-utility modules (`resolveRegion.mjs`, `matchRoute.mjs`, `dirtyState.mjs`). No subtle bugs introduced — I traced each helper's reference paths against the pre-PR `ThemeProviderHost.js` and they match line-for-line at the logic layer.

2. **`core:desktop` behavior change (silent WPDS fallback → NeutralProvider).** ✅ Acceptable. Desktop ships `compileStyles` for its `--wp-admin-shell--chrome--*` slot vocabulary and the engine CSS uses class-based selectors with hardcoded fallbacks — none of which depend on `WpdsThemeProvider` populating `--wpds-*` tokens. The one `--wpds-font-family-sans` reference (`engines/core-desktop/index.css:23`) has a robust 3-tier fallback. The behavior change is "loss of shell-level seed propagation into windows" which desktop never used. Documented in engine docstring + `docs/engines-and-design-systems.md` updates.

3. **No compat shim for `data-wpds-theme-provider-id`.** ✅ Correct. The attribute is internal-only — host emits it, host's own `buildScopedDetailCss` reads it, no external surface. A `grep` of `src/` + `tests/` shows the only remaining mentions are the historic-context line in the host docstring (acceptable) and the negative-assertion in the new test (explicit). External plugins/themes have no documented reason to target this attribute. Shipping a shim would just add cruft.

### Coverage gap analysis

The acknowledged JSDOM gap is real but bounded. Pure helpers cover everything the kernel can do without a React tree: density extraction, scope-content detection, selector routing, `compileStyles` integration, attribute-name constants. What's **uncovered** at unit level:

- **Engine `ThemeProvider` priority over `NeutralProvider`.** Only proven by manual smoke that each bundled shell renders with theming intact. Not in CI.
- **Error-boundary swap to `NeutralProvider` on render-time engine-provider throw.** No JSDOM = no React tree = no way to trigger `componentDidCatch`. Synthetic-test-only.
- **Density prop forwarding from `ProviderShell` → engine `ThemeProvider`.** The test asserts `pickDensity` returns the right value; doesn't assert the host passes it through to the inner provider. Manual smoke only.

This is consistent with the existing gap on `<PersistentRegion>`'s `data-app-mounted` attribute emission (also untested at JSDOM level). Acceptable for v3.0 beta given the precedent and the comprehensive manual smoke checklist in PR body. Worth filing a single "JSDOM scaffold" follow-up issue covering both gaps so they land together — or at least cross-link the two known-gaps in the eventual issue tracker.

### What's right

- The relocation is a true `git mv` — content-identical between `src/runtime/styles/WpdsThemeProvider.js` (deleted) and `src/runtime/engines/core-default/WpdsThemeProvider.js` (created). No functional drift smuggled in.
- The Slot/Fill audit is exhaustive: every consumer (the one `<Slot name="core:editor.sidebar">` in `simple-editor/index.js:405`, plus the transitive Slot/Fill use inside every `@wordpress/components` overlay component) mounts inside an app, which mounts inside a region, which mounts inside the engine `Layout`, which now wraps `<SlotFillProvider>`. `<NavigationGuard>` and `<BindingsConsumer>` — the two kernel-side components rendered outside engine Layout — render `null` and consume zero Slot/Fill substrate.
- The DOM attribute rename is thorough: emitted attribute on the wrapper, sibling `<style>` attribute, JSDoc on `EngineSource.ThemeProvider` contract, CLAUDE.md "Theming model" section, `docs/engines-and-design-systems.md`. The two negative paths (legacy attribute literal must NOT appear in build, must NOT appear in source) are both asserted by the new test (negative assertion in `buildScopedDetailCss` coverage).
- The error-boundary contract is now strictly "swap to NeutralProvider" — pre-PR it swapped to `WpdsThemeProvider`, which silently injected a DS-specific recovery path for a Material engine that just happened to ship a broken provider. Post-PR the recovery is DS-neutral and the failure mode is honest (no DS theming until the engine is fixed). This is the correct architectural choice for the kernel-pluggability contract.
- The `<style>` block emission preserves the engine's render-error-bubble contract: pure-compiler errors throw out of `buildScopedDetailCss` (asserted by test case 7) so the React error boundary catches them at React-render time and swaps to the neutral wrapper. Boundaries between sync errors (host catches) and async/runtime errors (engine handles) are clean.

1. **Expand `KERNEL_FILES` in `tests/runtime/kernel-no-ds-import.test.mjs`** to cover the 13 unlisted kernel files (or switch to filesystem auto-discovery). Mechanical change; closes a real contract-test gap.

2. **Widen `FORBIDDEN_PATTERNS`** to catch side-effect imports (`import '@wordpress/components/...'`), `require()` calls, and dynamic `import()`. The most likely regression vector — someone silences a CSS warning by side-effect-importing a layered stylesheet — slips through today's regex set.

3. **Reconcile test-count claim in PR body + CLAUDE.md.** Pick the counting convention and use it consistently. Off-by-121 doesn't undermine the PR but it's a paper trail risk.

The three above are all small mechanical edits; none are architectural. Once landed, this PR is in genuinely strong shape for merge.

dabowman:
## Review feedback addressed — three commits pushed

Thanks for the thorough review. All three top-priority items landed in granular commits on top of the existing branch.

### 1. 🟠 `KERNEL_FILES` allowlist incomplete — fixed via auto-discovery

Commit: `024eb0f`

Replaced the hardcoded 24-file array with a filesystem walker (`discoverKernelFiles()`) that walks `src/runtime/` recursively and excludes the `engines/` subtree wholesale. New kernel files inherit the DS-neutrality check automatically — manual lists drift by definition, as you noted.

Kernel file count: **24 → 37**. The 13 previously unlisted files now in coverage:
- `bindings/parseShortcut.mjs`, `bindings/triggerStore.mjs`
- `dirty-state/dirtyState.mjs`, `dirty-state/useDirtyState.js`
- `dashboardGrid/composeWidgets.mjs`
- `dataView/{deprecation,hydrateInline,mergeFields,useDataView}.{mjs,js}`
- `manifests/index.js`
- `modes/{resolveMode.mjs,useMode.js}`
- `registry/builtins.js`
- `shell-switching.js`
- `tokens/tokensResolver.mjs`

(No `lruCache.mjs` exists in the tree — flagged in your list but confirmed absent.)

### 2. 🟠 `FORBIDDEN_PATTERNS` only caught `from "..."` — fixed

Commit: `024eb0f` (same commit).

Replaced per-pattern regexes with a `makePattern( spec, matchSubpaths )` helper emitting one alternation covering all four import shapes:
1. `from "<spec>"` — named/default/namespace import
2. `import "<spec>"` — side-effect (CSS, etc.)
3. `import("<spec>")` — dynamic import expression
4. `require("<spec>")` — CommonJS (defensive)

Plus a pattern self-test block — positive samples (every shape, every package, including subpath imports like `@wordpress/components/build-style/style.css`) must match; negative samples (`@wordpress/element` / `@wordpress/i18n` / `@wordpress/compose` plus adjacent names like `@wordpress/components-extra` and prose-only comment mentions) must NOT match. Prevents the silent-regex-degradation failure mode where the patterns regress to catching nothing while every kernel file still emits PASS.

Verified end-to-end by temporarily injecting each forbidden shape into a kernel file:
- `import '@wordpress/components';` → FAIL detected ✅
- `const x = require('@wordpress/ui');` → FAIL detected ✅
- `const y = await import('@wordpress/icons');` → FAIL detected ✅
- `import '@wordpress/components/build-style/style.css';` → FAIL detected ✅

### 3. 🟡 `buildScopedDetailCss` hardcoded literal — fixed

Commit: `9ce38a9`

Hoisted both `THEME_SCOPE_ATTRIBUTE` + `THEME_SCOPE_DETAIL_ATTRIBUTE` constants from the bottom of `themeScope.mjs` to the top. `buildScopedDetailCss` now reads ``[${ THEME_SCOPE_ATTRIBUTE }="${ providerId }"]`` — single source of truth, future renames carry through automatically.

### 4. 🟡 Test-count claim reconciled — CLAUDE.md updated

Commit: `fde072e`

You're right that the PR's 750 doesn't survive the sum-of-`PASS:` convention. I verified the two counters explicitly:
- `npm run test:runtime 2>&1 | grep -E "^PASS:" | awk '{sum+=$2}'` → **629** (5 test files don't print a summary line)
- `for f in tests/runtime/*.test.mjs; do node "$f" | grep -cE "^PASS  "; done` → **750** before, **895** now

Sticking with the sum-of-individual-`PASS  ` convention (matches the historical CLAUDE.md counts). Updated to **895** runtime assertions; total now **1965**. Note: the bump from 750 → 895 includes the new pattern self-tests + the 13 extra kernel files × 7 patterns.

### Out-of-scope minor items (not addressed this round)

- 🟡 `core:desktop` chrome `--wpds-font-family-sans` reference — agree it's worth documenting, but out of scope for this PR per instructions (don't touch files outside `src/runtime/styles/` + `tests/runtime/` + CLAUDE.md). Filing as a follow-up.
- 🟡 No build-output assertion in CI — agreed, defer to a separate `tests/build/` follow-up.
- ⚪ Nits: `THEME_SCOPE_DETAIL_ATTRIBUTE` test coverage / `ThemeProviderHost.js:15` historical-context line / `core-single-pane` sibling-import comment / `source-types.js:96` JSDoc phrasing — noted; not changed this round to keep the diff tight.

### Test status

- `npm run test:runtime` → **895 / 895** ✅
- `npm run test:schema` → **103 / 103** ✅
- `npm run test:parity` → **4 / 4** ✅
- `npm run test:engines` → **7 / 7** ✅
- `npm run lint:js` → clean ✅
- `npm run lint:ts` → clean ✅
- `npm run build` → unchanged (220 KiB entry / 22 KiB CSS) ✅

dabowman:
## Re-review — PR #63 fixup pass

### Overview

Three fixup commits (`9ce38a9` / `024eb0f` / `fde072e`) addressed all three top-priority items from the original review. Verified end-to-end against the worktree at `fde072e`: the kernel-file walker auto-discovers 37 files (matching `git ls-tree` enumeration), the four-shape `makePattern()` catches every regression vector I probed, `buildScopedDetailCss` reads `THEME_SCOPE_ATTRIBUTE` via template interpolation (one literal remaining in the function is in JSDoc prose), and CLAUDE.md's 1965 / 895 totals reconcile with `npm run test:runtime` output. Build smoke clean at 220 KiB entry / 22 KiB CSS (matches PR claim, within rounding of original review's 216 KiB observation).

### Fixes verified

- **🟠 `KERNEL_FILES` auto-discovery (`024eb0f`).** `discoverKernelFiles()` walks `src/runtime/` recursively, excludes `engines/` by absolute-path comparison (`abs === enginesDir` short-circuit), filters to `.js|.mjs` only. Confirmed: `git ls-tree -r src/runtime/ | grep -E '\.(js|mjs)$' | grep -v '^src/runtime/engines/' | wc -l` = **37**, matching the walker's discovery and the PR claim. All 13 previously-unlisted files now in coverage. No non-JS/MJS kernel files exist (CSS/JSON only live under `engines/`), so the extension filter has no current false-positive risk.

- **🟠 `makePattern()` four-shape coverage (`024eb0f`).** Probed every shape I called out in the original review:
  - `import X from '@wordpress/components'` → caught ✅
  - `import { X } from '@wordpress/components'` → caught ✅
  - `import * as X from '@wordpress/components'` → caught ✅
  - `import '@wordpress/components/build-style/style.css'` (side-effect CSS — the originally-flagged regression vector) → caught ✅
  - `import('@wordpress/components')` (dynamic) → caught ✅
  - `require('@wordpress/components')` (CommonJS) → caught ✅
  - Multi-line `import {\n  foo,\n} from '@wordpress/components'` → caught ✅
  - `// uses @wordpress/components under the hood` (prose comment) → NOT matched ✅
  - `import { foo } from '@wordpress/components-extra'` (adjacent name) → NOT matched ✅
  - Allowed `@wordpress/element` / `@wordpress/i18n` / `@wordpress/compose` → NOT matched ✅
  
  Self-test corpus runs on every test execution (lines 218–329 of `kernel-no-ds-import.test.mjs`), guards against silent regex regression. Positive samples cover all 4 shapes for each of the 7 forbidden patterns; negative samples include adjacent package names + allowed packages + prose comments. The `(?:\/[^'"]*)?` subpath suffix is opt-in per pattern (off for the relative `WpdsThemeProvider` paths, on for `@wordpress/*` packages) — correct, since `./WpdsThemeProvider/foo` would not be a legitimate import shape.

- **🟡 `buildScopedDetailCss` consumes `THEME_SCOPE_ATTRIBUTE` (`9ce38a9`).** Both constants hoisted to the top (lines 26 + 33); `buildScopedDetailCss` reads via template interpolation (line 157: `` `[${ THEME_SCOPE_ATTRIBUTE }="${ providerId }"]` ``). Only remaining literal `'data-theme-scope-id'` is in the function JSDoc as a documentation example (line 101) — acceptable as prose. `theme-provider-host.test.mjs` still emits 40 assertions, no regression.

- **🟡 Assertion count reconciliation (`fde072e`).** Total: 1820 → 1965. Runtime: 750 → 895. `npm run test:runtime 2>&1 | grep -E '^PASS  ' | wc -l` returns **895** on my machine, matching the claim exactly. CLAUDE.md `kernel-no-ds-import` line updated 200 → 345. Verified by running the test file directly: `PASS: 345  FAIL: 0`. Math checks out: 37 files × 7 patterns = 259, plus 2 sanity checks, plus 28 positive-sample assertions, plus 56 negative-sample × pattern combinations = 345. Total math: PHP 886 + schema 103 + parity 4 + runtime 895 + engines 77 = 1965 ✓.

### 🔴 Blockers

None.

### 🟠 Major

None. Both Major items from the original review are closed.

### 🟡 Minor

None. The single Minor item I flagged (`buildScopedDetailCss` literal) is closed.

### ⚪ Nits

None on the fixup itself. The four nits from the original review (test coverage of `THEME_SCOPE_DETAIL_ATTRIBUTE` consumer, historical-context docstring line, single-pane sibling-import comment, `EngineSource.ThemeProvider` JSDoc phrasing) are intentionally not addressed this round per the response — fair call to keep the fixup diff tight. The three out-of-scope items (`core:desktop` font-family doc, build-output CI assertion, JSDOM scaffold) all belong in separate follow-ups; not blockers.

### Coverage gap analysis

Pure-helper coverage **expanded** in the fixup, not contracted:

- `theme-provider-host.test.mjs`: 40 assertions, unchanged. Covers `pickDensity` / `hasThemeContent` / `scopedSelector` / `buildScopedDetailCss` / attribute constants.
- `kernel-no-ds-import.test.mjs`: 200 → **345** assertions. Now covers (a) all 37 kernel files (was 24), (b) all 4 import shapes (was 1), (c) pattern self-tests (positive + negative samples).

Net: contract-test surface significantly stronger than at the original review. The static-analysis guard now legitimately codifies the kernel-DS-neutrality rule across the import shapes a future regression would actually use — the side-effect CSS import vector (originally my highest-probability concern) is now both pattern-caught AND self-tested. The acknowledged JSDOM gap (React-side `<ThemeProviderHost>` mount, engine `ThemeProvider` priority, error-boundary recovery, density-prop forwarding) is **unchanged in scope** — fixup didn't expand or contract it. Still adequate for v3.0 beta given the precedent set by the existing `<PersistentRegion>` `data-app-mounted` gap; a single follow-up issue covering both gaps would land them together. No newly-uncovered behavior introduced by the fixup.

### Regression hunting

- ✅ Walker correctly EXCLUDES `src/runtime/engines/**` (absolute-path short-circuit; verified by file count match).
- ✅ Walker correctly handles `.mjs` vs `.js` mix (regex test `/\.(?:js|mjs)$/` on entry name).
- ✅ Walker does NOT pick up `.css` / `.json` / `.md` files (extension filter excludes them; engines/ has the only CSS/JSON under src/runtime anyway).
- ✅ No symlink concerns (no symlinks in `src/runtime/`).
- ✅ `makePattern()` no false-positive on `@wordpress/components-extra` (anchored quote-char closing parenthesis).
- ✅ Pattern self-tests run on every execution, not just at module load — they're inline in the test body, between the file scan and Summary.
- ✅ Build smoke: `npm run build` clean at 220 KiB entry / 22 KiB CSS, matches PR claim.
- ✅ Full runtime suite: 895 assertions, 0 failures.

### Verdict

**Approve.** All three fixup commits landed cleanly, faithfully addressed the original review's findings, and meaningfully strengthened the contract test surface. No regressions introduced. Ship it.

PR #68, opened 2026-05-27
feat(nav): make engine `menu-renderer` field functional

dabowman:
## What

The `menu-renderer` engine field was **decorative**. `admin-engine.json` described it (`sidebar-drilldown` / `sidebar-tree` / `dock` / `drawer` / `none` / `plugin:*`) plus a `wp_admin_shell_register_menu_renderer()` plugin contract, but `grep menu-renderer src/ includes/` returned zero readers — menu rendering was hardcoded in `core:navigation`. This implements the mechanism (rather than downgrading the field to advisory-only).

An engine now **names** a menu-rendering strategy; `core:navigation` **dispatches** to the registered renderer. Built-in, engine-owned, and `plugin:*` renderers all resolve through one registry — the seam a non-WPDS engine plugs into without touching kernel code.

## How

- **Kernel registry** — `src/runtime/config/menuRendererRegistry.js`, DS-neutral, mirrors `iconMap` (register/resolve by id, first-write-wins, dev-warn fallback). Holds opaque component refs, imports nothing DS.
- **Compile thread** — `buildRuntimeConfig` stamps the active engine's `menu-renderer` onto the runtime config.
- **`core:navigation` → dispatcher** — orders + prunes the `menu` tree once, resolves the renderer the engine named, renders `<Renderer items currentPrimary navConfig />`. `none` → renders nothing; absent → falls back to `sidebar-drilldown`.
- **Shared pure helpers** — `src/runtime/menu/menuTree.mjs` (order/prune/walk), node-importable, `window`-free, so every renderer shares one implementation. `pruneMenu` takes an injected `passes` predicate; the `userCan`-backed `itemPassesPermissions` stays host-side.
- **Renderers** — drilldown moved to `_renderers/SidebarDrilldownRenderer.js`; **new** `SidebarTreeRenderer.js` (expandable in-place tree, auto-expands active ancestors).
- **Engine-owned drawer** — `engines/core-single-pane/DrawerRenderer.js` (collapsible accordion) self-registers from the single-pane **engine module**, so it travels with the engine on extraction. Depends only on kernel + WPDS, never on the nav app's internals.
- **Engine declarations** — `core:default` → `sidebar-drilldown`, `core:single-pane` → `drawer`, `core:desktop` → `dock`.
- **PHP entry point** — `wp_admin_shell_register_menu_renderer( $renderer_id, $args )` (`WP_Admin_Shell_Menu_Renderers`): validates `plugin:{slug}/{name}` ids (core ids reserved), requires a `script` handle, enqueues it on the shell page. `window.wpAdminShell.registerMenuRenderer` published in `src/index.js`.

| `menu-renderer` | Renderer | Owner |
|---|---|---|
| `sidebar-drilldown` (default) | slide-in sub-screens + `?screen=` slot; honors `config.collapsed` | bundled `core:navigation` |
| `sidebar-tree` | expandable in-place tree | bundled `core:navigation` |
| `drawer` | accordion sections | `core:single-pane` engine module |
| `dock` | dock-rail tiles | `core:desktop-dock-app` (not via `core:navigation` — declaration is intent-only) |
| `none` | renders nothing | — |
| `plugin:{slug}/{name}` | plugin-supplied | third-party plugin |

## Behavior change

⚠️ **`core:single-pane` now renders a real accordion drawer** instead of reusing the drilldown sidebar. Worth a manual smoke before tagging.

## Tests

- New `tests/runtime/menu-renderer-registry.test.mjs` + `menu-tree.test.mjs`.
- Extended `build-runtime-config.test.mjs` (stamping) + `run-manifest-tests.php` (`register_menu_renderer` success / reserved-id / bad-id / missing-script / dup).
- Green: `test:runtime`, `test:schema` (56), `run-manifest-tests.php` (76), `npm run build`, lint.

## Known gap (flagged, not solved here)

The kernel mounts synchronously, so a **loose plugin renderer script** enqueued after the bundle can race the first paint — built-in + engine-owned renderers register via direct ESM import and are race-free. Robust loose-script support needs a published kernel import surface; this is the same blocker for the planned `core:single-pane` / `core:desktop` → standalone-plugin extraction. Logged in `docs/feedback.md` with three fix options.

## Not included

`CLAUDE.md` Navigation + extension-surface-#15 edits were left in the working tree — entangled with an unrelated in-flight doc refactor (Pipeline/Theming condensing + extraction to `docs/engines-and-design-systems.md` / `docs/code-map.md`). To be folded into that refactor's commit.


dabowman (commented):
## Review — solid, one a11y fix before merge

Sound refactor. `menu-renderer` was dead schema; now wired through a kernel registry (mirrors `iconMap`), `core:navigation` → thin dispatcher, pure tree helpers extracted to `menuTree.mjs`. Verified locally: `menu-renderer-registry`, `menu-tree`, `build-runtime-config`, and **`kernel-no-ds-import`** all green — the new kernel files (`menuRendererRegistry.js`, `menuTree.mjs`) import nothing DS, neutrality intact.

### 🔴 Should-fix — double ARIA landmark

`SidebarTreeRenderer.js` and `core-single-pane/DrawerRenderer.js` both wrap output in `<nav aria-label="Navigation">`. But they mount inside regions that already declare `role: "navigation"`:

- `core-single-pane/engine.json` — `core:nav-drawer` → `role: "navigation"`, mounts `core:navigation` → `DrawerRenderer`. **Confirmed nested navigation landmark.**
- `sidebar-tree` mounts in the same navigation-role sidebar slot → same doubling.

Violates the documented rule: *"Apps don't add their own ARIA landmark element. Region wrappers already render `role={region.role}`… NavigationApp dropped its outer `<nav>` for this reason."* The moved `SidebarDrilldownRenderer` correctly omits `<nav>` — the two new renderers are inconsistent with it.

**Fix:** swap `<nav aria-label=…>` → plain `<div className=…>` in both. The region already supplies the landmark.

### 🟡 Nits

- **DRY:** `subtreeContainsPrimary` is duplicated byte-for-byte in `DrawerRenderer` + `SidebarTreeRenderer`. It's pure + `window`-free → belongs in `menuTree.mjs` next to the other walk helpers.
- **PR body path drift:** says `includes/cascade/class-wp-admin-shell-menu-renderers.php`; actual path is `includes/manifests/`.
- `SidebarTreeRenderer` ignores `navConfig` entirely (no title/description header, unlike drilldown). Likely intentional for the tree idiom — flagging, not blocking.

### 🟢 Good

- First-write-wins consistent JS+PHP; an engine can't be clobbered by a plugin.
- Engine-owned drawer self-registers from the single-pane module (eager-imported in `builtins.js`) → race-free, travels on extraction.
- Dispatcher fallback chain (`unknown → default`, `none → null`) never blanks the nav.
- Loose-script race honestly scoped + logged, not hand-waved.
- PHP entry point validates `plugin:{slug}/{name}`, reserves core ids, requires a script handle, dup-rejects.

Double-landmark is the only thing I'd gate merge on.

dabowman:
Addressed in `43563b8`:

- **🔴 Double nav landmark** — `SidebarTreeRenderer` + `DrawerRenderer` no longer wrap in `<nav aria-label>`; both now render a plain `<div>`, matching the rule the drilldown renderer follows (region already supplies `role="navigation"`). Added a code comment at each site noting why.
- **🟡 DRY** — `subtreeContainsPrimary` moved to `src/runtime/menu/menuTree.mjs` (pure, `window`-free) next to the other walk helpers; both renderers import it. Local copies deleted. New test case in `menu-tree.test.mjs`.
- **🟡 PR body path** — current body has no `includes/cascade` reference; the path is `includes/manifests/class-wp-admin-shell-menu-renderers.php` throughout.
- **🟡 `SidebarTreeRenderer` ignores `navConfig`** — intentional for the tree idiom (no root title/description header); left as-is per your flag.

Verified: `test:runtime` (incl. `menu-tree` + `kernel-no-ds-import`), lint, `npm run build` all green.

dabowman (commented):
## Re-review — both findings resolved ✅

Commit `43563b8` addresses the review cleanly.

- **Double landmark** — `SidebarTreeRenderer` + `DrawerRenderer` swapped `<nav aria-label>` → plain `<div>`, with comments noting the region already supplies `role="navigation"`. Now consistent with the drilldown renderer. Stray `__` imports removed.
- **DRY** — `subtreeContainsPrimary` lifted into `menuTree.mjs` (pure, `window`-free), both renderers import it. New JSDoc distinguishes it from `findContainerForPrimary`. Added `menu-tree.test.mjs` coverage (direct child / deep descendant / no-match / empty-primary / leaf).

Verified locally: `menu-tree` suite green (20 passed incl. the new depth cases), lint clean on both renderers.

No remaining blockers. The PR-body path typo (`includes/cascade/` → `includes/manifests/`) is cosmetic. LGTM.

PR #87, opened 2026-05-27
Add customizable definition to admin.json schema

### LINKED ISSUE #80: admin.json schema has no `customizable` definition
From `docs/feedback.md` (2026-05-05, V2.M4 review). Verified still open against the current canonical schema.

`docs/schemas/admin.json` declares no `customizable` (nor legacy `userCustomizable`) `$defs` definition, and none of the entry types (`screen`, `menuItem`, `command`, `workspaceWidget`, …) include the property. The field is fully implemented:
- PHP cascade: `includes/cascade/class-wp-admin-shell-customizable.php` (`const FIELD = 'customizable'`, semantics: boolean / array-of-paths / absent=locked).
- Runtime: `compileStyles` `NON_TOKEN_KEYS`, `AppearanceApp`.

So hand-authored shells can't validate per-entry `customizable` declarations against the schema (IDE schema lookup gives authors no signal).

**Fix:** add a `$defs/customizable` definition and reference it from the entry types that honor it.

*Note: the related "LEGACY_FIELD = 'userCustomizable' lifetime unbounded" item is obsolete — no such constant exists in the current `WP_Admin_Shell_Customizable`; the legacy fallback was removed.*

dabowman:
## Summary

The `customizable` field is fully implemented in the PHP cascade (`WP_Admin_Shell_Customizable`) and the runtime (`compileStyles` `NON_TOKEN_KEYS`, `AppearanceApp`), but `docs/schemas/admin.json` declared no `$defs/customizable` definition and none of the entry types referenced it. Hand-authored shells therefore got no IDE/Ajv validation for per-entry `customizable` declarations.

This adds a `$defs/customizable` definition and references it from every entry type that genuinely honors the field.

## Shape

`$defs/customizable` is a `oneOf` of:
- `boolean` — `true` (entry fully writable by consumer origins) / `false` (locked)
- `array` of non-empty `string` dotted paths, `uniqueItems: true` — only the listed paths are writable

Absence is documented as equivalent to `false` (default-deny — same posture as block supports). This matches `WP_Admin_Shell_Customizable::read_decl` / `filter_writes` exactly: `true` → all; `null`/`false`/non-array → locked; array → string dot-paths only. The description also records the hardcoded deny-list (`screens.*.permissions`, `screens.*.app`, `commands.*.invoke`, `workspace.engine`) that blocks consumer-origin writes even with a matching allowlist entry.

## Entry types that gained the property

Derived from the PHP enforcement surfaces (`V3_TOP_LEVEL_BLOCKS` + `filter_doc`'s styles handling + per-entry `read_decl`):

- `screen`
- `menuItem`
- `command`
- `workspaceWidget`
- `styles`
- `region`
- `route`

(`settings.applications`/`settings.regions` referenced by the legacy `filter_settings` path are not part of the v3 `settings` block — v3 moved per-app/per-region overrides under `styles.applications`/`styles.regions`, already covered by `styles`' `additionalProperties: true` — so no schema change there. The `userCustomizable` legacy fallback is obsolete and was not added.)

## Validation

`npm run test:schema` passes — 59 PASS / 0 FAIL, including every bundled shell. No bundled shell currently declares `customizable`, so the change is purely additive with no validation regression.

Closes #80

https://claude.ai/code/session_01NGYDqHhZjMnMgaxvsmgpEq

---
_Generated by [Claude Code](https://claude.ai/code/session_01NGYDqHhZjMnMgaxvsmgpEq)_

### DIFF
--- a/docs/public/admin-json-reference.md
+++ b/docs/public/admin-json-reference.md
@@ -24,6 +24,7 @@ This reference covers the admin.json workspace schema (`admin.json`).
 - [styles](#styles)
 - [preload](#preload)
 - [regions / routes](#regions--routes-escape-hatches)
+- [customizable](#customizable)
 - [v2 surfaces (deprecated)](#v2-surfaces-deprecated)
 
 ## JSON Schema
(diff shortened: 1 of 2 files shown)


dabowman:
## Review: Approve with nits

Verified the PR branch (not the local base checkout). JSON is well-formed (`json.tool` clean) and the `$defs/customizable` shape + the 7 `$ref` placements are correct.

**Semantics vs PHP (`WP_Admin_Shell_Customizable`) — match, in the safe direction:**
- `true` → all writable, `array` → listed dotted paths only, `false`/absent → locked. Matches `filter_writes`/`filter_subtree`/`path_is_allowed`.
- Schema's `oneOf` (boolean | array) is *stricter* than PHP, which also tolerates object/number/null and treats them as locked. Rejecting those at author time is correct and the description documents it explicitly ("object/number/other types are treated as locked", "absence is equivalent to false"). No security gap — schema can only be more restrictive than enforcement.
- `uniqueItems`/`minLength: 1` add constraints PHP doesn't require (PHP `array_filter(..., 'is_string')` drops non-strings, dupes are idempotent) — harmless tightening.
- Array allowlist match is exact-tail (`tail_matches_any`), not prefix; descriptions say "the listed dotted paths" which is accurate.

**$ref placement — all 7 land in the right `$defs`:** `screen` (366), `menuItem` (493), `command` (530), `workspaceWidget` (561), `styles` (764), `region` (898), `route` (1060). `styles` root-level matches PHP's `read_decl($upstream['styles'])` path; `settings.applications`/`settings.regions` entries (also enforced via `filter_settings`) are `additionalProperties: true`, so no ref needed there. `userCustomizable` correctly absent (repo-wide grep: none).

**Nits (non-blocking):**
- `workspace` itself is a V3 top-level block PHP's `path_is_allowed` walks from the root, and the `workspace` object is `additionalProperties: false` with no `customizable` property — so an author can't legally write `workspace.customizable` even though PHP would honor it. Likely intentional (workspace-level lock is rarely wanted and `workspace.engine` is deny-listed anyway), but worth a one-line note if it's deliberate.
- The new descriptions point authors to `docs/public/admin-json-reference.md`, but that file doesn't currently document `customizable`. Consider adding a section there (or dropping the cross-ref until it exists).

Description prose matches the file's inline-documentation house style. `npm run test:schema` not re-run here (no `node_modules` in this checkout) — relying on the reported 59/0; the change is ref-only additions to existing valid defs, so a pass is expected.

---

PR #88, opened 2026-05-27
Plan alpha 0.1.0: workspace as primary admin, file-based override

dabowman:
## Summary

This PR documents the plan for alpha release 0.1.0, which establishes the workspace as the primary admin entry point and introduces `wp-content/admin.json` as a partial override file (rather than a full shell replacement). The plan supersedes the archived 2026-05-21 draft and locks three key architecture decisions.

## Key changes

- **Baseline + override model**: `wp-admin-default` moves into the `core` origin as the shipped baseline; `wp-content/admin.json` becomes a partial override in the `plugin` slot, deep-merging on top (theme.json model).
- **Workspace as primary entry**: `/wp-admin/`, `/wp-admin/index.php`, and `/wp-admin/admin.php` render the workspace when active (file present or legacy option set). The legacy `?page=wp-admin-shell` route is removed.
- **Classic escape hatch**: Toolbar button (cap-gated to `manage_options`) sets a session cookie to temporarily revert to classic wp-admin; reciprocal "Back to workspace" link in classic admin bar.
- **Link interception (both directions)**: Workspace clicks on `/wp-admin/...` links route through the admin-route registry or iframe-fallback; classic links to mapped screens redirect into the workspace.
- **Partial-permissive validation**: Override files relax top-level `required` fields; completeness enforced post-merge by shape-tests.
- **Endpoint allowlist**: Hard-coded paths (`admin-ajax.php`, `customize.php`, network admin, etc.) never hijack; extensible via filter.

## Workstream breakdown

Nine workstreams (~7 days serial; W4/W5 can run parallel after W2):

1. **W1** (~1.5d): Core baseline + file override origin — move `wp-admin-default` to `includes/origins/admin-default.json`, create `Origin_File` for `wp-content/admin.json`, teach validator partial-permissive mode.
2. **W2** (~1.5d): Workspace as primary admin entry — `admin_init` hijack at priority 0, render shell on allowlisted paths, remove legacy settings page.
3. **W3** (~0.5d): Classic escape hatch — toolbar button + session cookie + reciprocal "Back to workspace" link.
4. **W4** (~1d): Workspace → classic link interception — capture-phase listener, registry lookup, iframe-fallback fallthrough.
5. **W5** (~0.5d): Classic → workspace redirect — `legacy_path` + `legacy_query` route metadata, mapper in hijack.
6. **W6** (~1d): Chromeless bridge generalization — promote iframe-internal link bridge from `core:desktop-iframe` to kernel-neutral platform service.
7. **W7** (~0.5d): Redirect-loop guards + endpoint allowlist — centralize never-hijack paths, audit boundary cases.
8. **W8** (~1d): Tests + readiness doc — wire new PHP/JS test runners, author `docs/alpha-readiness.md`.
9. **W9** (~0.5d): Documentation + release notes — update README, CLAUDE.md, design spec, schema sketch, feedback log; set version to `0.1.0`.

## Notable implementation details

- **Cascade change**: The resolver's baseline-skip rule inverts — keep baseline + merge when `plugin` slot holds a partial override file; skip baseline only on back-compat path when `plugin` slot holds a complete selected shell. Detect via origin source, not by sniffing for engine key.
- **Engine resolution**: Partial override file may omit `workspace.engine`; effective engine reads `override.workspace.engine ?? baseline.workspace.engine`.
- **Workspace active detection**: `wp_admin_shell_workspace_active()` returns true when (a) valid `wp-content/admin.json` present, OR (b) `wp_admin_shell_active_shell` option explicitly written to DB (back-compat). Fresh install has neither → no hijack.
- **Core-portability**: Layout mirrors future WordPress core feature (`wp-includes/admin.json`

https://claude.ai/code/session_01LsZ725nKYd7QWGyCmpdjhH


dabowman:
## Code review — workspace hijack, classic-mode, file override, link interception

Reviewed the implementation (not the plan doc). 14 findings, most severe first. The two **HIGH** items produce silent, user-visible breakage on standard configs and are the ones I'd block on.

### Correctness

**1. [HIGH] Classic-mode cookie path hardcoded `/wp-admin/` — escape hatch breaks on subdirectory/relocated installs**
`includes/class-wp-admin-shell-classic-mode.php:101`
The cookie is set with `'path' => '/wp-admin/'`, but `admin_url()` (used for the toggle link) is `/blog/wp-admin/` on a subdirectory install, multisite subdirectory site, or relocated wp-admin. The browser scopes the cookie to `/wp-admin/`, so it's **never sent** on `/blog/wp-admin/*` requests → `passes_base_gates()` never sees it → the hijack always re-fires. The admin can never reach classic wp-admin, and with the settings page removed there's no alternate off-switch. Use `ADMIN_COOKIE_PATH` (or derive from `parse_url( admin_url('/'), PHP_URL_PATH )`).

**2. [HIGH] Legacy redirect silently drops nonce-protected classic GET actions**
`includes/class-wp-admin-shell-hijack.php:190` (`maybe_redirect_legacy` / `match_legacy_hash`)
Score-0 screens (no `legacy_query`: `edit.php`, `upload.php`, `users.php`, `plugins.php`, `themes.php`, `edit-comments.php`, `tools.php`…) match **any** GET to that script vacuously. A direct navigation to e.g. `plugins.php?action=deactivate&plugin=foo&_wpnonce=abc` or `edit-comments.php?action=spamcomment&c=10&_wpnonce=…` is 302-redirected to the workspace route, dropping `action`/`_wpnonce` — the action never executes and the admin gets no indication. The JS `classifyAdminLink`/`matchLegacyRoute` has the same flaw. Fix: skip the redirect when `action`/`_wpnonce` are present (or only redirect bare-script URLs).

**3. [MED-HIGH] JS `matchLegacyRoute` lacks the unresolved-token guard PHP has**
`src/runtime/navigation/adminLinkInterceptor.mjs:122`
PHP `match_legacy_hash` skips entries with unresolved tokens (`false !== strpos($interpolated, '{')` + `''`/`'/'` checks). JS does not. Clicking `post.php?action=edit` with no `post` param matches `/posts/{id}/edit` (legacy_query satisfied), `query.get('post')` is `null` → the literal token survives → `navigate('/posts/{id}/edit')`. An empty `?post=` interpolates to `''` → `navigate('/posts//edit')`. Either way the SPA lands on a bogus route. Mirror the PHP guard.

**4. [MED] iframe bridge forwards every non-`route` classification into `iframe.src`**
`src/runtime/platform/iframeBridge.mjs:56-61`
`classifyBridgeMessage` only special-cases `action === 'route'`; everything else — RPC endpoints, the `?classic=1/0` toggle, cross-origin, in-page `#` hashes (all `'pass'` at the top level) — falls through to `{ type:'iframe', href:url }`, so the parent sets `iframe.src` to it. A classic-mode toggle clicked inside an embedded page loads classic *inside* the iframe instead of escaping; the URL reaches `iframe.src` as an unvalidated postMessage sink. Only forward `action === 'iframe'`; treat `'pass'` as ignore.

**5. [MED, latent] iframe bridge origin pin silently disabled when `adminUrl` is relative**
`src/runtime/platform/iframeBridge.mjs:100-104`
`new URL('/wp-admin/')` throws → `expectedOrigin = null` → the origin check never runs (iframe-fallback falls back to `'/wp-admin/'` when `window.wpAdminShell.adminUrl` is unset). The `event.source` pin saves the wired case, but any consumer omitting `getIframeWindow` loses both pins. Refuse all messages when `adminUrl` can't be parsed to an origin.

### Parity / lower severity

**6. [LOW] PHP/JS interpolation encoder mismatch** — `rawurlencode` (`hijack.php:258`) vs `encodeURIComponent` (`adminLinkInterceptor.mjs:126`) differ for `!*'()`; the server redirect and client interception produce different router params for the same link. Numeric IDs unaffected; free-text tokens drift.

**7. [LOW] `String(query.get(key))` coerces an absent param to `'null'`** — `adminLinkInterceptor.mjs:109`. A `legacy_query` value of literal `'null'` spuriously matches a link omitting the param; empty-string constraints diverge from PHP's `'' === $got`. Latent. Use `query.has(key)`.

**8. [LOW] `?classic=0` is an unauthenticated GET state-flip (CSRF)** — `classic-mode.php:60`. Clears the cookie with no nonce/cap; a crafted top-level navigation can bounce an admin out of classic. Low impact (UI-state cookie, no privileged write); `?classic=1` is cap-gated.

**9. [LOW] Override file read has no size/depth cap** — `origin-file.php` (`load()`). `file_get_contents` + default-depth `json_decode` on a filter-controllable path, on cache-cold admin requests. Site-author-controlled, so hardening — add a `filesize()` guard.

**10. [LOW] admin.json cache signal is mtime-only (1s resolution)** — `cache.php:94`. A malformed file fixed within the same filesystem second keeps the same mtime → stale baseline-only tree served until another invalidation. Mix in filesize or a content hash.

**11. [LOW] `is_root_entry()` takes over bare `admin.php` dispatch** — `hijack.php:521`. Plugins dispatching on `admin.php` with `action=` but no `page=` get short-circuited by `render_and_exit()` while a workspace is active. Rare pattern.

**12. [LOW, efficiency] Full resolver runs on every non-root classic admin GET** — `hijack.php:206`. `maybe_redirect_legacy` calls `wp_admin_shell_get_active_config()` + rebuilds `legacy_map` on every classic page that clears the base gates, even when nothing maps (cached after first resolve).

### Cleanup

**13. [LOW] Dead settings-page references** — `classic-menu-bridge.php:145` (skip-list `'wp-admin-shell-settings'`/`'wp-admin-shell'`) + stale rationale comment at `:50`. Slugs are now unreachable; if a plugin later claims the vacated `wp-admin-shell` slug it'd be silently skipped from ingestion.

**14. [LOW] Orphaned `wp_admin_shell_settings` settings group** — `wp-admin-shell.php:718`. The `register_setting` group + its justifying comment survive, but the `options.php` form that rendered it (`wp_admin_shell_render_settings`) was removed; switch surfaces now POST to `/wp/v2/settings`.

---

**Verified clean:** fresh-install detection (`workspace_active()` returns false — the option's registered default is `''`); the `$menu` seeding in `render_and_exit` is a guarded no-op (WP builds `$GLOBALS['menu']` on `admin_menu`, before `admin_init`); partial-override deep-merge + null tombstones work (plugin slot uses `merge_authoritative`); `engine_origin()`'s string signature has no array-passing callers; `legacy_*` keys survive the cascade; no redirect loop between hijack/toggle/legacy-redirect.

<sub>🤖 Generated with [Claude Code](https://claude.ai/code)</sub>

---

dabowman:
Thanks for the review — disposition below. Fixes pushed in `0caaec2`.

**Fixed**

- **#1 [HIGH] cookie path** → `ADMIN_COOKIE_PATH` (covers subdir / relocated / multisite-subdir; hardcoded `/wp-admin/` is gone).
- **#2 [HIGH] nonce-protected GETs** → both the PHP redirect (`match_legacy_hash`) and JS `classifyAdminLink` now bail when `_wpnonce` is present, so `plugins.php?action=deactivate&_wpnonce=…` etc. complete in classic. Guarded on `_wpnonce` specifically (not `action`) so the `post.php?action=edit` → `/posts/{id}/edit` mapping survives.
- **#3 [MED-HIGH] JS unresolved-token guard** → `matchLegacyRoute` now skips a candidate whose interpolation leaves a `{token}`, is empty/`/`, or collapses to `//` — mirrors the PHP guard. Also adopted `query.has()`/empty-string semantics (**#7**) for PHP parity.
- **#4 [MED] iframe bridge sink** → `classifyBridgeMessage` only forwards a genuine same-origin admin **miss** to `iframe.src`; a `'pass'` (RPC / classic toggle / cross-origin / hash) is now ignored.
- **#5 [MED] origin pin** → resolves a relative `adminUrl` against the page origin and **refuses all messages** when no origin can be determined (no more silent pin-disable).
- **#9 / #10 [LOW] override file** → 1 MB size cap before read; cache signal is now `mtime:size` so a same-second malformed→fixed edit invalidates.
- **#11 [LOW]** → `is_root_entry()` no longer takes over `admin.php?action=…` plugin dispatch.
- **#13 [LOW]** → dropped the dead `wp-admin-shell-settings` / `wp-admin-shell` skip-list entries (+ stale comment).

**Deferred, with rationale**

- **#6** rawurlencode vs encodeURIComponent — numeric IDs (the only tokens core maps) are identical; free-text drift is latent. Left a note rather than diverge the encoders.
- **#8** `?classic=0` CSRF — it flips a UI-state cookie only (no privileged write; `?classic=1` is cap-gated). Adding a nonce to the admin-bar link is more friction than the risk warrants for alpha.
- **#12** resolver on classic GETs — `wp_admin_shell_get_active_config()` is resolver-cached after the first run, so it's a cache hit, not a re-resolve.
- **#14** `wp_admin_shell_settings` `register_setting` — kept: it still provides the option's `show_in_rest` + sanitize callback (the switch surfaces POST to `/wp/v2/settings`). Only the now-unused group **name** is cosmetic; removing the registration would break the REST write.

Tests: interceptor 45, iframe-bridge 17, routing runner gains nonce + unresolved-token cases; `lint:js` + all node suites green. The "verified clean" items match my read too. Happy to revisit any of the deferred four if you'd rather close them now.

---

dabowman:
## Code Review — Workspace as primary admin, file-based override

Reviewed the full diff (plan doc **and** the shipped implementation). Quality is high overall — strong security instincts (origin+source pinning on postMessage, `_wpnonce` passthrough, RPC allowlist, cap-gated escape hatch) and good pure-logic test coverage. Two real bugs below.

---

### 🔴 Bug 1 — bare `edit.php` mapping hijacks every custom post type (correctness, high)

The baseline `posts` screen got `legacy_path: "edit.php"` with **no** `legacy_query`:

```jsonc
"posts": { ..., "legacy_path": "edit.php" }      // shells/wp-admin-default.json
```

`matchLegacyRoute` / `match_legacy_hash` treat a zero-constraint entry as a vacuous match (score 0). So a workspace-active GET to:

- `edit.php?post_type=product` (WooCommerce)
- `edit.php?post_type=acme_event` (any CPT plugin)

→ 302-redirects into the workspace `/posts` list. **Every third-party CPT list screen becomes unreachable** in workspace mode — it lands on the core Posts list instead.

**Why the tests don't catch it:** both fixtures (`run-alpha-routing-tests.php` `$map`, `admin-link-interceptor.test.mjs` `ROUTES`) give `/posts` an explicit `legacy_query: { post_type: post }` and assert `edit.php?post_type=book → null`. The *shipped baseline* has no such constraint, so the green tests describe a config that isn't what ships.

**Fix tension:** adding `legacy_query: {post_type: post}` to the baseline posts screen fixes CPTs but then bare `edit.php` (no `post_type`, = posts in classic) stops matching. The equality model can't express "absent OR equal to `post`". Options:

- Treat a missing `post_type` as `post` in the matcher (special-case the default), **or**
- Only let a zero-constraint `edit.php` entry match when the URL also has no `post_type` — fall through (no redirect, classic renders) for unknown CPTs.

Fall-through is safest: an unmapped CPT staying classic is correct. Either way, **add a fixture with the real baseline shape** (`edit.php?post_type=product`, posts entry exactly as shipped) so the regression is pinned.

---

### 🟡 Bug 2 — PHP/JS parity gap: `//` guard missing PHP-side (low-med)

JS `matchLegacyRoute` rejects a degenerate interpolation:

```js
if ( '' === interpolated || '/' === interpolated ||
     interpolated.includes( '{' ) || interpolated.includes( '//' ) ) continue;
```

PHP `match_legacy_hash` omits the `//` check:

```php
if ( '' === $interpolated || '/' === $interpolated || false !== strpos( $interpolated, '{' ) ) { continue; }
```

A route like `/x/{a}/{b}` with one token resolving empty yields `/x//7` — JS skips, PHP redirects to it. The comment claims "Mirrors the JS interceptor's matchLegacyRoute"; make them actually match — add `false !== strpos( $interpolated, '//' )`.

---

### 🟢 Minor notes

- **Perf:** `maybe_redirect_legacy` calls `wp_admin_shell_get_active_config()` (full resolve) on every non-root admin GET that clears base gates — including every plugin `?page=` page, which never matches. Resolver cache softens it, but the legacy map is built and discarded on each such load. Consider a cheap "is this pagenow even a candidate" pre-check.
- **`legacy_map` collision:** the screens loop builds `$map[path]`, then the programmatic-registry loop does `$map[$path] = …` unconditionally — a registered route sharing a screen's path silently clobbers the screen entry. Last-write-wins is probably fine but undocumented.
- **`match_legacy_hash` tie-break:** two zero-constraint entries on the same `legacy_path` resolve by map iteration order. Unlikely in practice; worth a comment.
- **Plan vs. impl naming drift:** plan says `adminLinkInterceptor.js` / `iframeBridge.js`; impl ships `AdminLinkInterceptor.js` (React effect) + `adminLinkInterceptor.mjs` (pure) + `iframeBridge.mjs`. The split is *better* (follows the `.mjs`-for-node-testable convention) — just flagging the plan doc is now stale on filenames.
- **Baseline not relocated:** plan W1 said move to `includes/origins/admin-default.json`; impl keeps it at `shells/wp-admin-default.json` (`BASELINE_SLUG`). Harmless — `shells_mtime()` already covers the cache signal — but the plan's "extend `core_mtime`" step is now moot and the core-portability note still implies relocation.

---

### Security — looks right

- postMessage handler pins both `event.origin` (against `adminUrl`, refuses when origin can't be resolved) and `event.source` (bound iframe window). Spoof tests present. ✅
- `classifyBridgeMessage` refuses to feed a `'pass'` URL (RPC / classic-toggle) into `iframe.src` — only a genuine same-origin admin miss becomes an iframe nav. ✅
- `_wpnonce` passthrough in both interceptor and PHP mapper prevents dropping nonce'd destructive actions. ✅
- `?classic=1` activation cap-gated to `manage_options`; 1 MB size guard + `is_assoc` gate on the override file. ✅
- Static-file exposure of `wp-content/admin.json` acknowledged + deferred (structural config, no secrets). Fine for alpha; keep it visible in install docs.

### Tests

Good pure-logic coverage (interceptor + bridge classification, partial-permissive load, tombstone, mtime cache signal). Gaps: the CPT case (Bug 1), and the render/redirect `exit` paths (reasonably deferred to manual smoke since `is_admin()` is false under WP-CLI).

---

**Verdict:** Solid, well-documented work. Suggest **blocking on Bug 1** — it breaks CPT admin for any plugin (WooCommerce included) and the test suite currently hides it. Bug 2 + the parity comment are quick follow-ups; the rest is polish.

dabowman:
## Re-review — after `0caaec2`

Nice turnaround. The hardening commit lands well:

- ✅ Classic-mode cookie scoped to `ADMIN_COOKIE_PATH` (subdir / relocated / multisite-subdir installs).
- ✅ `_wpnonce`-protected requests never mapped, both PHP redirect and JS interceptor — `plugins.php?action=deactivate&_wpnonce=…` completes in classic.
- ✅ iframe bridge: only a genuine same-origin admin MISS reaches `iframe.src`; `'pass'` is ignored, not piped in as a sink. Refuses all messages when origin can't be resolved.
- ✅ JS `matchLegacyRoute` mirrors the unresolved-token / `//` guard; absent params compare as `''` via `query.has()`.
- ✅ 1 MB override cap + `mtime:size` cache signal; `is_root_entry()` no longer grabs `admin.php?action=…`.

The skip rationale (#6 / #8 / #12 / #14) is reasonable.

Two items from my review are **still open**, though — neither was in scope of the nonce work:

### 🔴 Bug 1 — bare `edit.php` still hijacks every CPT (unaddressed)

`shells/wp-admin-default.json` (posts screen) still ships:

```jsonc
"legacy_path": "edit.php"      // no legacy_query
```

The `_wpnonce` guard doesn't cover this — a CPT **list** browse is a plain GET with no nonce:

- `edit.php?post_type=product` (WooCommerce) → vacuous score-0 match on the posts entry → 302 to `/posts`.

So any third-party CPT list screen is unreachable in workspace mode (lands on core Posts). This is the blocker from the first pass; the nonce fix is orthogonal to it. Needs either a `post_type`-aware matcher (missing `post_type` ⇒ `post`) or a zero-constraint `edit.php` entry that only matches when no `post_type` is present (fall through → classic). Plus a fixture using the **shipped** posts shape (`edit.php?post_type=product` → no redirect).

### 🟡 Bug 2 — PHP `match_legacy_hash` still lacks the `//` guard

JS got the guard; PHP didn't. `includes/class-wp-admin-shell-hijack.php`:

```php
if ( '' === $interpolated || '/' === $interpolated || false !== strpos( $interpolated, '{' ) ) {
```

Add `|| false !== strpos( $interpolated, '//' )` so a multi-token route with one empty segment (`/x//7`) is rejected the same way both sides — the comment already claims parity with the JS matcher.

Everything else from the first review is either fixed or fairly skipped. Bug 1 is the one I'd still block on.

dabowman:
Both open items fixed in `4b5e6ec`.

**🔴 Bug 1 — CPT hijack** — fixed two ways together:
- Baseline `posts` and `posts-new` now declare `legacy_query: { post_type: "post" }`, so `edit.php?post_type=product` (any CPT) matches neither `/posts` nor `/pages` → falls through to classic. ✅
- The matcher (JS `matchLegacyRoute` **and** PHP `match_legacy_hash`) treats an **absent** `post_type` as `post` (WP convention: bare `edit.php` / `post-new.php` == `post_type=post`), so a bare `edit.php` still maps to `/posts` despite the new constraint.

Net behavior: `edit.php`→`/posts`, `?post_type=post`→`/posts`, `?post_type=page`→`/pages`, `?post_type=product`→classic. Added fixtures with the **shipped** posts shape (`admin-link-interceptor.test.mjs` `SHIPPED`, plus the CPT case in `run-alpha-routing-tests.php`) so the regression is pinned, not hidden by a more-constrained test fixture.

**🟡 Bug 2 — PHP `//` guard** — added `|| false !== strpos( $interpolated, '//' )` to `match_legacy_hash`; the two matchers now actually agree as the comment claimed.

Also picked up two of the 🟢 notes since they were cheap: `maybe_redirect_legacy` now early-returns for plugin `?page=` pages (and `_wpnonce`'d requests) before the resolve+map build, and `legacy_map` documents its last-write-wins / iteration-order tie-break.

Tests: interceptor 49, iframe-bridge 17, schema 59; `lint:js` + all node suites green. PHP suites still need a wp-env run (no CI on the repo) — the render/redirect `exit` paths remain manual-smoke per `docs/alpha-readiness.md`.

---

dabowman:
## Code review — PR 88 (ultra, recall mode)

9 finder angles × verify × sweep. PR ships real implementation (hijack, classic-mode, origin-file, resolver, link interceptor, iframe bridge), not just the plan doc. **15 findings**, ranked. Most CONFIRMED with quoted lines.

### 🔴 High — workspace takeover / auth boundary

**1. `is_root_entry()` hijacks `index.php?page=<plugin-subpage>`** — `includes/class-wp-admin-shell-hijack.php` (~L130)
`if ( 'index.php' === $pagenow ) { return true; }` returns true unconditionally — no `$_GET['page']` guard. The `admin.php` branch right below correctly guards `empty($_GET['page'])`. Any plugin dashboard subpage via `add_dashboard_page()` / `add_submenu_page('index.php', …)` (e.g. `/wp-admin/index.php?page=my-tool`) gets `render_and_exit()`'d into the workspace instead of reaching its handler. Mirror the `admin.php` guard.

**2. classic→workspace redirect ignores per-screen permissions** — `class-wp-admin-shell-hijack.php` (`passes_base_gates` L115 / `maybe_redirect_legacy` L193-224)
Only floor is `current_user_can('read')`; the target screen's `permissions` block is never consulted before the 302. Subscriber typing `/wp-admin/plugins.php` (mapped at `shells/wp-admin-default.json:955`) gets redirected to `#/plugins` instead of WP's native cap deny. Workspace gates the screen later, so not direct escalation — but the hard classic deny becomes a soft in-workspace gate.

**3. `?classic` toggle has no allowlist guard, runs on exempt endpoints** — `includes/class-wp-admin-shell-classic-mode.php` (`handle_toggle`, `admin_init` pri -10)
Only guard is `isset($_GET['classic'])`. No `is_admin()`/ajax/REST/ENDPOINT_ALLOWLIST check that the hijack itself carefully applies. A stray `?classic=0` on `update.php`/`customize.php`/`plugin-install.php` (all fire `admin_init`) triggers `set_cookie(false)` + `wp_safe_redirect` + `exit`, short-circuiting a request the hijack deliberately exempts.

**4. `?classic=0` is CSRF-able + cookie checked by presence only** — `classic-mode.php` (`handle_toggle`) + `hijack.php` (`passes_base_gates` L107)
No nonce anywhere (two `phpcs:ignore NonceVerification` pragmas). Activation (`=1`) is `manage_options`-gated; deactivation (`=0`) is open to any logged-in user — `// Anyone may return to the workspace.` So `` flips an admin out of classic via CSRF. And `passes_base_gates` bails on `!empty($_COOKIE['wp_admin_shell_classic'])` — value never validated, so any non-empty forged/garbage cookie permanently disables the workspace for that browser.

### 🟠 Medium — wrong routing / silent failures

**5. Editing a Page opens the Posts editor** — `shells/wp-admin-default.json` (`post-edit` L713 / `page-edit` L810)
`post-edit` maps `post.php?action=edit` → `/posts/{id}/edit` (config `postType:post`). `page-edit` has NO `legacy_path` block. Classic `post.php?post=42&action=edit` carries no `post_type`, and neither matcher resolves the post object — so every `post.php?action=edit` (Pages, any CPT) funnels into the Posts editor. Both JS interceptor and PHP redirect agree on the wrong target. `post.php` is post-type-agnostic; the mapping can't disambiguate by query alone.

**6. `set_cookie()` silent no-op when headers already sent** — `classic-mode.php` (`set_cookie` / `redirect_without_param`)
`setcookie()` wrapped in `!headers_sent()`, but `$_COOKIE` is mutated in-memory unconditionally and `wp_safe_redirect()+exit` fires regardless. If output already started: cookie never persists, redirect fails to send (headers warning), browser stays on `?classic=` URL — toggle silently no-ops client-side while server thinks it succeeded.

**7. Override file gets full trusted-tier cascade authority** — `includes/cascade/class-wp-admin-shell-resolver.php` (L197-237)
`wp-content/admin.json` loads into the `plugin` slot (in `TRUSTED_ORIGINS`) and merges via `merge_authoritative` — skips `Customizable::filter_doc` + `Permissions::enforce_origin_tier`. So whoever can write the file (compromised theme/plugin, low-priv FTP) can delete baseline screens via null tombstones, **grow** `screens[].permissions` OR-sets, and change `workspace.engine` — same authority as the bundled plugin, bypassing the shrink-only rules for site/role/user. Confirm this is intended; if so, document the trust assumption.

**8. `is_valid_partial()` does no structural validation** — `includes/origins/class-wp-admin-shell-origin-file.php` (L148)
Only gate is non-empty assoc array (`is_assoc`). A screen missing `id`/`path`, malformed `permissions`, etc. passes, merges into the plugin slot, and corrupts the resolved tree at runtime. Completeness is only caught post-merge by `run-shape-tests.php` in CI — a live site degrades rather than falling back to the bare baseline.

**9. "Switch shell" is a silent no-op when `wp-content/admin.json` exists** — `class-wp-admin-shell-resolver.php` (file branch wins, L190-202) + `src/runtime/shell-switching.js` + user-menu
File override wins over `wp_admin_shell_active_shell` (by design). But the switcher writes the option + reloads, and nothing consults `Origin_File::exists_and_valid()` to hide/disable it. User picks a shell, page reloads, nothing changes, no diagnostic.

**10. iframe bridge `external-link` has no scheme allowlist (reachable in prod)** — `src/runtime/platform/iframeBridge.mjs` (L70-72, L146-152) via `src/apps/iframe-fallback/index.js:57`
`classifyBridgeMessage` returns `{type:'external', href:url}` for any non-empty string; `installIframeBridge` calls `w.open(href, …)`. `iframe-fallback` installs the bridge **without** `openExternal`, so it hits the default `w.open`. Bridge fires for ANY same-origin iframe-mounted admin page (`Sec-Fetch-Dest: iframe`), so a stored-XSS/compromised embed can post `url:'data:…'`/`'javascript:…'`. Origin pin still applies and modern browsers block `javascript:` in `window.open`, so defense-in-depth gap — add an `^https?:`/`mailto:` allowlist like the admin-link path already does.

**11. iframe bridge source-pin is optional** — `iframeBridge.mjs` (L86, L97, L127-133)
`event.source` check wrapped in `if (typeof getIframeWindow === 'function')`; param is documented optional. Omit it → only `event.origin` checked → any same-origin window (sibling frame, popup) can spoof admin-link/external-link and drive navigation, contradicting the module's "source-pinned" docblock. Prod caller (iframe-fallback) passes it; gap is for future callers + untested path.

**12. iframe bridge drops re-auth / error signals** — `iframeBridge.mjs` (L74) + `iframe-fallback`
Forwards only `admin-link`/`external-link`; every other PHP chromeless-bridge type (`reauth-detected`, `iframe-error`, …) returns `{type:'ignore'}`. Generic iframe-fallback embed never surfaces re-auth — session expires inside the frame, user sees a blank/broken iframe with no prompt.

### 🟡 Lower — latent / edge

**13. Cache signal misses equal-size same-second edits** — `class-wp-admin-shell-origin-file.php` (`signal()` ~L130)
Signal is `mtime:size` (1s `filemtime` resolution). An edit of identical byte length within the same wall-clock second (swap a hex color, equal-length label) produces an identical signal → resolver serves the stale tree until the 300s TTL. Docblock claims size "disambiguates an edit made within the same filesystem second" — only for size-changing edits. Use a content hash. New trigger test only covers presence/absence, never an equal-size edit.

**14. `classifyAdminLink` throws on relative `adminUrl`, disabling ALL interception** — `src/runtime/navigation/adminLinkInterceptor.mjs` (L183-190) + `AdminLinkInterceptor.js:24`
`new URL(adminUrl)` with no base throws on a relative string; catch returns `{action:'pass'}` for every link. Default fallback is `'/wp-admin/'` (relative). `iframeBridge.mjs:112-117` guards this exact case with `new URL(adminUrl, w.location.href)`; this file doesn't. Latent: WP `admin_url()` is absolute in prod, so only the defensive fallback path triggers it — but then interception is silently dead.

**15. Nonce-less `?action=` GET can be redirected, dropping the action** — `class-wp-admin-shell-hijack.php` (`maybe_redirect_legacy` L212 / `match_legacy_hash` L242)
Only POST and `isset($_GET['_wpnonce'])` are excluded. A nonce-less state-changing `?action=` GET on a mapped `legacy_path` whose `legacy_query` doesn't include `action` gets 302'd to the workspace, dropping `action` and voiding the op. No bundled-shell config obviously triggers it (core gates state changes behind nonces), but plugin-contributed `legacy_path` + nonce-less action GET would.

---

### Also noted (cleanup / not bugs)
- **Empty `{}` file doesn't activate workspace** (`origin-file.php` L149 `empty()` reject) — intentional per docblock, but contradicts the "file present ⇒ active" prose in README/alpha-readiness. No diagnostic without `WP_DEBUG`.
- **Orphaned `register_setting('wp_admin_shell_settings', …)`** (`wp-admin-shell.php:718`) — its only consumer (the deleted settings page) is gone; option now writable via REST only. Dead registration.
- **JS/PHP duplication**: `matchLegacyRoute` (JS) ↔ `match_legacy_hash` (PHP), and `RPC_SCRIPTS` (Set) ↔ `ENDPOINT_ALLOWLIST` (already drifted) — must edit in lockstep across two languages.
- **`encodeURIComponent` (JS) vs `rawurlencode` (PHP)** in token interpolation — diverge on `! ' ( ) *` (not spaces). Latent; shipped token values are integers.
- **Perf on hot path**: `Origin_File` re-stats the file 3× per request (`load`/`signal`/`mtime`); `mtime()` is a dead accessor (cache uses `signal()`); `legacy_map()` not memoized (rebuilt each redirect). Memoize on the admin_init pri-0 path.
- **Ingested plugin screens get no `legacy_path`** — classic↔workspace round-trip only works for the ~40 hardcoded core screens; Yoast/Woo/ACF pages surface in nav but don't round-trip.

dabowman:
Ultra-review addressed in `53308bc` — 11 of 15 fixed, 4 deferred with rationale.

**Fixed**
- **#1** `is_root_entry()` now guards the `index.php` branch too — `index.php?page=…` (`add_dashboard_page()` subpages) and `?action=` dispatch reach their handler instead of being hijacked.
- **#3** classic-mode toggle bails on ajax/REST/cron/xmlrpc/CLI + non-GET, so a stray `?classic=` can't short-circuit an exempt endpoint.
- **#4 (cookie)** `passes_base_gates` matches the exact value `'1'` — a garbage/forged non-empty cookie can't permanently disable the workspace.
- **#5** dropped the `post.php` mapping from the baseline `post-edit` screen. `post.php` carries no `post_type`, so it can't disambiguate Pages/CPTs — editing a Page no longer opens the Posts editor; it stays classic.
- **#8** structural gate rejects a wrong-typed known top-level block (`"screens":"oops"`) → falls back to baseline instead of corrupting the tree.
- **#9** shell switcher hides + `switchShell()` throws when a `wp-content/admin.json` override is active (new `window.wpAdminShell.workspaceFileActive` flag) — no more silent no-op.
- **#10** iframe-bridge `external-link` is scheme-allowlisted (`http`/`https`/`mailto`); `javascript:`/`data:` are ignored, never handed to `window.open`.
- **#11** iframe-bridge source-pin is now mandatory — messages refused when no `getIframeWindow` is provided (was origin-only, spoofable by sibling frames).
- **#14** `AdminLinkInterceptor` passes an absolute `adminUrl`, so the relative fallback can't throw and silently disable all interception.
- **#15** both matchers skip an entry that doesn't itself constrain `action` when the URL carries one — a nonce-less state-changing `?action=` GET isn't redirected with `action` dropped.
- **#7 / #13** documented the intended trusted-tier authority of the override file; corrected the cache-signal docblock to not overclaim.

**Deferred, with rationale**
- **#2** classic→workspace redirect ignoring per-screen perms — the workspace's own 4-layer capability gating prunes/denies the target screen downstream, so it's a hard-deny→soft-deny UX nuance, not escalation (your review noted this). Duplicating permission eval into the redirect path is a design choice I'd rather get your call on before adding.
- **#4 (CSRF nonce on `?classic=0`)** — flips a UI-state cookie only (no privileged write; `=1` is cap-gated). The full fix spans JS+PHP nonce plumbing for a low-impact toggle; happy to add if you want it for alpha.
- **#6** `set_cookie` on `headers_sent()` — the toggle runs at `admin_init` pri -10, before any admin output, so headers aren't sent on the real path; the guard is just defensive.
- **#12** bridge dropping `reauth-detected`/error signals — a feature add (re-auth UX for generic embeds), not a regression; the desktop engine keeps its own handling.

On the cleanup notes: kept `register_setting` (still needed for the option's `show_in_rest`); left the JS↔PHP matcher duplication + `encodeURIComponent`/`rawurlencode` drift as documented (numeric IDs only); the ingested-plugin-screens-have-no-`legacy_path` limitation is already in `alpha-readiness.md`.

Tests: interceptor 51, iframe-bridge 21 (scheme + mandatory-source-pin + action-guard cases), routing runner gains the `index.php?page` guard, trigger runner gains the bad-structure fixture. schema 59; `lint:js` + all node suites green. PHP suites still need a wp-env run (no CI on the repo).

---

dabowman:
Things I'm noticing from testing: 

If you move the admin.json out of wp-content after it was there before you get this persistent error in wp-admin. 

Warning: filesize(): stat failed for /wordpress/wp-content/admin.json in /wordpress/wp-content/plugins/wp-admin-shell/includes/origins/class-wp-admin-shell-origin-file.php on line 90
Warning: file_get_contents(/wordpress/wp-content/admin.json): Failed to open stream: No such file or directory in /wordpress/wp-content/plugins/wp-admin-shell/includes/origins/class-wp-admin-shell-origin-file.php on line 97
Warning: Cannot modify header information - headers already sent by (output started at /wordpress/wp-content/plugins/wp-admin-shell/includes/origins/class-wp-admin-shell-origin-file.php:90) in /wordpress/wp-includes/functions.php on line 7220
Warning: Cannot modify header information - headers already sent by (output started at /wordpress/wp-content/plugins/wp-admin-shell/includes/origins/class-wp-admin-shell-origin-file.php:90) in /wordpress/wp-admin/admin-header.php on line 14

the error persists even after deactivating/adtivating the admin shell plugin. 

I would expect removing admin.json to just revert to the classic wp-admin. Maybe with some sort of notification, but not a persistent error. We need to consider this state after a user removes the config file.

dabowman:
Good catch — that was a real bug, fixed in `6ceb7b2`.

The chain: `is_readable( $path )` was returning **true for the deleted file** (PHP's stat / realpath cache holding a stale "exists" entry across requests — common on Docker bind-mount filesystems), so `filesize()` and `file_get_contents()` ran on a path that wasn't actually there and emitted warnings. Those warnings became premature output, which broke the `header()` calls in admin-header.php → the cascading "headers already sent" errors. The plugin deactivate / reactivate didn't help because reactivating runs the same code on the next request.

Fixed defensively in all three stat-using methods (`load`, `mtime`, `signal`):

- `clearstatcache( true, $path )` immediately before the check, so we don't trust the OS / PHP cache for this specific file path.
- Added `is_file()` to belt-and-suspenders against a directory path the `wp_admin_shell_admin_json_path` filter might return.
- `@`-suppressed the read calls (`filesize` / `filemtime` / `file_get_contents`) and continue to handle their `false` returns — so any race between the check and the read can't leak a warning to output.

Net behavior: removing `wp-content/admin.json` reverts cleanly to classic wp-admin on the next request — no warnings, no header-already-sent cascade, no manual flush needed. Will send the rebuilt zip in the next turn so you can drop it in.

---

dabowman:
noticing that if the user adds a screen to admin.json that already exists in the default, it duplicates it. For example, I added the profile screen as a top level menu item. It still kept the profile item in the user sub-menu. We should de-dupe this somehow. by path? by app? not sure exactly how but it's confusing to add a pre-existing screen and still retain it's default location. Here's the admin.json file I'm using for testing: 

```
{
	"$schema": "plugins/wp-admin-shell/docs/schemas/admin.json",
	"version": 3,
	"$wpds": "6.9",
	"name": "smoke-test",
	"title": "Smoke Test Shell",
	"description": "Minimal admin.json override. Presence of this file at wp-content/admin.json triggers the shell takeover; the cascade resolver layers this delta over the bundled wp-admin-default baseline.",
	"workspace": {
		"engine": "core:default",
		"default-screen": "home",
		"branding": {
			"title": "Smoke Test Shell"
		}
	},
	"screens": {
		"home": {
			"label": "Home",
			"icon": "admin-home",
			"path": "/",
			"app": "core:dashboard",
			"permissions": {
				"capabilities": [ "read" ]
			}
		},
		"profile": {
			"label": "Profile",
			"icon": "user",
			"path": "/profile",
			"app": "core:profile",
			"permissions": {
				"capabilities": [ "read" ]
			}
		}
	},
	"menu": {
		"home":    { "position": 10 },
		"profile": { "position": 20 }
	},
	"styles": {
		"theme": {
			"color": {
				"bg": "#f6f7f7"
			}
		}
	}
}

```

dabowman:
Fixed in `62ae50a` — implemented as **shallowest-wins dedupe** in `bind_screens`.

The trade-off you flagged (by path? by app?) — I went with **by screen id, keeping the shallowest occurrence in the menu tree**. Specifically:

1. `bind_screens` (the `wp_admin_shell_data` priority-5 callback) now does a pre-pass that walks the merged menu tree and records the minimum depth at which each screen-bound id appears.
2. A second walk drops any entry whose id appears at a shallower depth elsewhere.

For your test file: baseline has `menu.users.items.profile` (depth 2); your override declares `menu.profile` (depth 0). Min-depth for `profile` is 0 → the nested copy at depth 2 is dropped, the top-level one stays. The Users container itself is preserved (it has other children); only the duplicate Profile child inside it goes away.

**Why this rule.** "Author moved it to a higher level" maps cleanly to "shallowest wins" — top-level placement most clearly reflects intent. By-path or by-app would also dedupe correctly here but would also drop legitimate alternates (e.g. an author who genuinely wants Profile in two places). By-screen-id only fires when the **same id** binds in two spots; manual menu items with distinct ids + an explicit `href` are never touched, so the escape hatch for "I really want two surfaces for the same screen" is: `menu.profile-shortcut: { label: "My Profile", href: "#/profile" }` under whichever parent you like.

The user-menu **app** (the avatar dropdown) is separate from the `menu` tree and unaffected — it reads its own data, not the cascade menu. That's intentional: it's an always-available quick-access surface.

Verified standalone against your test file plus three regression cases (no-override-preserves-nested, distinct-id-not-deduped, container-preserved). CLAUDE.md notes the behavior. New zip incoming next turn.

---

dabowman:
noticing that if a session expires while the shell is open the re-login flow is odd. I had the framed add a plugin screen open, restarted the site and then reloaded the page. The iframe reloaded but then prompted be to authenticate inside the iframe. I did eventually get the login popup modal at the shell level, but this is a weird state where you can get it in the iframe before you get it in the shell.

dabowman:
with iframes it can be slow to hide the chrome sometimes it will show the chrome for a few seconds before hiding. We should show a loading state until the page is totally loaded and we've hidden the chrome. Right now we just see loading until it loads and there's a flash of chrome.

dabowman:
iframe flow to upload a plugin get's stuck if it's replacing a current plugin. The buttons to replace or cancel don't respond leaving the user stranded. Here's the html of the buttons: 

<p class="update-from-upload-actions">
<a class="button button-primary update-from-upload-overwrite" href="update.php?action=upload-plugin&amp;package=74&amp;overwrite=update-plugin&amp;_wpnonce=4017805307" target="_parent">Replace current with uploaded</a> 
<a class="button" href="update.php?action=upload-plugin-cancel-overwrite&amp;package=74&amp;_wpnonce=e32af98fc1">Cancel and go back</a>
</p>

dabowman:
All three iframe-fallback complaints fixed together in `37f56e9` (they're all the same render path):

**1. Chrome flash → gone.** The iframe is now rendered `visibility: hidden` from the start; the Spinner stays up until `onIframeLoad` has injected the chrome-hide CSS, *then* the iframe is revealed. So the user never sees a moment of un-hidden wp-admin chrome. Navigation to a different src (or a re-fetch) resets visibility too, so this also covers the in-shell navigation case.

**2. Session expiry → no more in-iframe login form.** When `onIframeLoad` detects the loaded document is the WP login page (`/wp-login.php`, a `#loginform` element, or `body.login` class), the iframe stays hidden AND `window.wp.heartbeat.connectNow()` is called immediately — the shell-level `wp-auth-check` modal pops at once instead of waiting up to ~15s for the next scheduled poll. A `heartbeat-tick` listener watches for the `wp-auth-check` false→true transition and reloads the iframe so it re-fetches the real page once you've re-authenticated.

**3. Stuck "Replace current with uploaded" button.** The button has `target="_parent"` — the WP plugin-install flow wants the parent frame to navigate to `update.php?action=…&_wpnonce=…` (allowlisted to classic). The in-iframe chromeless bridge's link interceptor (sub-system 9) was `preventDefault`'ing the click and posting `wp-admin-shell-admin-link` up to the parent, where the new `_wpnonce` guard then ignored it → button looked broken. The interceptor now respects `target="_parent"` and `target="_top"` and lets the browser do the natural navigation. The classic update.php handles the upload via the allowlist, then redirects you back.

Sending the rebuilt zip next turn.

---

dabowman:
theme thumbnails aren't loading in the /themes dataview. here's the rendered html. No images being rendered as  or as background-image with css. not sure which is expected but no image is there in the DOM.  

<div class="dataviews-view-grid__media dataviews-view-grid__media--placeholder"><span class="dataviews-view-grid__media-placeholder"></span>::after</div>

dabowman:
Fixed in `f4a4038`. Two things going on, and I want to flag the diagnostic gap so you can confirm.

**The fix.** When `record.screenshot` is empty/missing from the `/wp/v2/themes` REST response, the renderer now falls back to the canonical `<siteUrl>/wp-content/themes/<stylesheet>/screenshot.png` path. Browsers silently drop a 404, so the worst case stays "no image visible" — but for any wordpress.org-shipped default theme (Twenty Twenty-Five, etc.) the file lives at exactly that path, so it resolves. Extracted as `screenshotUrl(item)` so the grid renderer and the details modal preview share it.

**Why it was empty.** WordPress's `WP_Theme::get_screenshot()` returns the URL **only if** the theme directory contains `screenshot.{png|gif|jpg|jpeg|webp}` — otherwise `''`. The REST endpoint forwards that as `record.screenshot`. A few scenarios produce empty there:
- Slim Playground / wp-now theme bundles that strip the asset.
- Themes that don't ship a screenshot at all (rare for wordpress.org themes, common for custom ones).
- REST cache-buster or context issue (less likely — your fetch uses `context: 'edit'` which exposes the field).

If you want to nail down which one, in the browser console with the themes view loaded:
```js
( await wp.apiFetch( { path: '/wp/v2/themes?context=edit&status=active,inactive' } ) )
  .map( t => ({ slug: t.stylesheet, screenshot: t.screenshot }) )
```
If the URLs are present in that response, the fallback didn't trigger and there was a different issue (e.g. core-data normalization stripping the field). If they're empty/missing, the fix above is the right call and you should see images now.

New zip incoming.

---

dabowman:
iFramed settings screens like privacy are loading nested instances of the shell. so privacy loads an infinite cascade of shell windows.

dabowman:
when I upload a new plugin I still see the flash of chrome after I go through the upload dialog and before I am presented with the replace/cancel state for an existing plugin. When I choose replace it uploads successfully but lands me on the classic screen. If I reload I get the shell again.

dabowman:
while data views are loading they show the empty no results state before loading the items. It should have a loading state instead of an empty state while items are loading.

Issue #75, opened 2026-05-27
Entity-CRUD apps: screen-spec feature gaps (status counts, quick/bulk edit, directory browse, etc.)


dabowman:
From `docs/feedback.md` (2026-05-26, DataViews/DataForm standardization review). The standardization pass was deliberately behavior-preserving; `docs/screens/*.md` specify capabilities the current apps don't implement. Tracked here as a roadmap bundle.

- **Posts / Comments / Users / Plugins** — status-tab counts (need parallel count requests or a custom aggregate endpoint).
- **Posts** — Quick Edit / Bulk Edit + hierarchical (pages) tree.
- **Taxonomy** — split-pane add form + hierarchical tree + default-category protection.
- **Users** — role counts, email-change confirmation, application-passwords, bulk role-change.
- **Comments** — inline reply, quick-edit, Mine/Pings tabs, undo snackbar.
- **Plugins / Themes** — Add-New directory browse, install progress, auto-update toggle, paused/dependencies state (most lack a REST endpoint — need custom routes).
- **Media** — list mode, image-edit canvas, type counts.

Source of truth for each: the matching `docs/screens/*.md` functional spec.

dabowman:
Audit 2026-06-04 — the **wave-2 integration (PR #243) landed several items**; updating the checklist. Implemented on `main`:

- ✅ Posts — **Bulk Edit** (modal) + **status-tab counts** (PR #234)
- ✅ Users — **role counts** + **bulk role-change**
- ✅ Comments — **inline reply** (reply modal) + edit modal + status tabs/verbs (PR #232)
- ✅ Media — **list (table) mode** alongside grid

Still open:

- Posts — Quick Edit; hierarchical (pages) tree
- Taxonomy — split-pane add form; hierarchical tree; default-category protection
- Users — email-change confirmation; application-passwords (#108); 
- Comments — quick-edit; Mine/**Pings** tabs; undo snackbar
- Plugins / Themes — Add-New directory browse; install progress; auto-update toggle (#147 upstream); paused/dependencies state
- Media — image-edit canvas (#125); type counts
- Status-tab counts for Comments/Users/Plugins (posts only so far)

Source of truth per item remains the matching `docs/screens/*.md` spec.

dabowman:
@claude validate and address in a new PR

PR #90, opened 2026-05-27
feat(dataviews): status/role/type counts on entity-CRUD filters (#75)

dabowman:
## Summary

Addresses the **status-count** slice of #75 — surfacing per-value counts on the entity-CRUD filters, the shell's equivalent of wp-admin's status tabs (`Published (12)`, `Editor (3)`, `Images (40)`).

DataViews has no native count slot on filter elements, so counts are folded into the element **labels** (the filter `value` is never touched, so filtering keeps working). Counts are **global** — they ignore the active search/page so they mirror wp-admin's status links rather than re-fetching on every keystroke — and inherit the list's capability gating (a value the user can't read simply omits its count instead of rendering a misleading zero).

### What changed

- **New shared hook** `src/apps/_shared/dataviews/useEntityElementCounts.js` — fires one lightweight `per_page=1&_fields=id` request per value through a single `useSelect` over `core-data`, reading totals off the `X-WP-Total` header (via `getEntityRecordsTotalItems`).
- **New pure helper** `withElementCounts` + an `elementCounts` option on `buildFields` (`buildFields.mjs`) — merges `{ value: count }` into a field's resolved `elements`, working for both spec-declared and fallback elements.
- **Wired into the apps:**
  - **Posts / Comments** — status counts (known enum from `STATUS_LABELS`).
  - **Users** — role counts, with role values pulled from the resolved spec `roles` elements so the set tracks whatever roles the shell exposes.
  - **Plugins** — status counts derived **client-side** from the existing one-shot fetch (no extra requests).
  - **Media** — media-type counts on the `SelectControl`, plus an unfiltered total for `All`; reuses `withElementCounts` even though Media isn't a DataViews app.
- **Docs** — updated each app's `app.json#documentation` (`data.reads`) + `app.md` (architecture / known-limitations parity notes).
- **Tests** — pinned `withElementCounts` and the `elementCounts` option in `tests/runtime/dataviews-shared.test.mjs`.

### Scope note

#75 is a roadmap bundle spanning many features (quick/bulk edit, directory browse, image editing, hierarchy trees, etc.). This PR intentionally covers only the **status/role/type counts** item across Posts, Comments, Users, Plugins, and Media. The remaining items stay tracked on the issue.

Counts render as **filter-dropdown options**, not the standalone subsubsub tab strip (`All | Mine | …`) wp-admin uses, and there is no `Mine` count — noted in each app's `app.md`.

## Test plan

- [x] `npm run test:runtime` — green (incl. new `withElementCounts` / `elementCounts` cases; 30/30 in the shared suite)
- [x] `npm run test:schema` — green (59/59; app.json edits validate)
- [x] `npm run lint:js` on all touched files — clean
- [x] `npm run build` — succeeds (only pre-existing bundle-size warnings)
- [ ] Manual: open Posts/Comments/Users/Plugins, confirm filter dropdown shows live counts; confirm Media type dropdown shows counts incl. `All`
- [ ] Manual: cap-gating — a subscriber sees no misleading counts on admin-only statuses

https://claude.ai/code/session_016wZAgkFgT5Zn2GtQK9QJfD

---
_Generated by [Claude Code](https://claude.ai/code/session_016wZAgkFgT5Zn2GtQK9QJfD)_


dabowman:
## Code review

Traced the new `useEntityElementCounts` hook + `withElementCounts` helper through all five apps, the core-data cache semantics, and the REST query shapes. Solid, well-tested PR — the `buildFields` refactor preserves prior behavior, the Comments `approved`/`hold` and Users `roles` count values are valid query params for their endpoints, and the `_fields`-stripped stable key means the partial count queries don't truncate the main list's shared cache bucket.

One correctness issue worth fixing before merge, plus two efficiency notes.

### 1. Counts go stale after mutations (Posts / Comments / Users / Media) — should fix

After trash/delete/status-change, each app invalidates only the **list** query:

```js
invalidateResolution( 'getEntityRecords', [ 'root', 'comment', queryArgs ] );
```

But the count requests use different args — `{ [field]: value, per_page: 1, _fields: 'id', context: 'edit' }` (and Media's `allCount` uses `{ per_page: 1, _fields: 'id', context: 'edit' }`). None of those are invalidated, so the filter labels keep showing pre-mutation totals.

- Approve a pending comment → `Pending (3)` / `Approved (12)` stay `3` / `12` instead of `2` / `13`.
- Trash a post → `Published (12)` stays `12`; `Trash (n)` doesn't grow.
- Delete an image → `Images (12)` / `All (40)` stay stale.

core-data caches resolution across SPA navigation, so the wrong numbers persist for the whole session — only a full page reload refreshes them. Comments is the worst case since status-change *is* its primary workflow. (Plugins is exempt — it derives counts from the same `records` that already get invalidated.)

**Fix:** also invalidate the count queries (or the entity's `getEntityRecords` resolver broadly) in each app's `onSettled`.

### 2. Eager per-value count fan-out on every mount — efficiency

`useEntityElementCounts` fires one REST request per value unconditionally at mount, even if the filter is never opened: Posts 6 (incl. trash/future/private), Comments 4, Users one-per-role, Media 5 (incl. `allCount`). wp-admin gets every status count from a single aggregate query (`wp_count_posts`/`wp_count_comments`). Partly inherent to the "global counts" design (no single-request REST equivalent exists), so flagging as a known cost rather than a defect — worth a note in `docs/perf-baseline.md` if cold-mount request count matters.

### 3. Media `allCount` is redundant when unfiltered — minor

When `mediaType === ''`, the main `useEntityRecords('root','media', { per_page: 40, page, context: 'edit' })` already exposes the unfiltered `totalItems` (same context+stableKey bucket); the dedicated `allCountQuery` duplicates that fetch. Could read the main list's `totalItems` when unfiltered and only fall back to a dedicated query when a type filter is active.

---

dabowman:
## Re-review of `b0448fd`

The fixup cleanly addresses all three findings.

### 1. Stale counts after mutations — fixed

New `invalidateEntityElementCounts(invalidateResolution, kind, name, field, values, baseQuery)` in `src/apps/_shared/dataviews/useEntityElementCounts.js` mirrors the hook's query shape exactly:

```js
{ ...baseQuery, [ field ]: value, per_page: 1, _fields: 'id', context: 'edit' }
```

Byte-identical to what `useEntityElementCounts` emits, so the resolver cache keys line up and `invalidateResolution` actually clears them. Wired in:

- **Posts** — trash modal `onSettled` → `STATUS_VALUES`.
- **Comments** — status-change `onSettled` and trash modal `onSettled` → `STATUS_VALUES`.
- **Users** — delete modal `onSettled` → `roleValues` (correctly added to the actions `useMemo` deps).
- **Media** — upload and delete → `MEDIA_TYPE_VALUES` + dedicated `ALL_COUNT_QUERY` invalidation.
- **Plugins** — correctly left alone (counts derive from `records`, refreshed by the existing list invalidation).

Callback dep arrays check out — module-level constants don't need listing, and `roleValues` (component-scoped memo) is added where it's referenced.

### 2. Eager per-value count fan-out — documented

`docs/perf-baseline.md` now carries the per-app cold-mount request count table, names the wp-admin parity gap (`wp_count_posts` / `wp_count_comments` have no REST aggregate equivalent), and notes the bounded shape (smallest payload, global counts, post-mutation re-fetch). Honest framing — known cost, not a defect.

### 3. Media `allCount` redundancy — fixed

When `mediaType === ''`, `allCount` reuses `mainTotal` from the main list's `useEntityRecords`. A dedicated `useSelect` fires only when a type filter is active (where the main list's total is scoped). Saves one request on the common path.

### Minor nits — optional

- The new `invalidateEntityElementCounts` helper isn't pinned by a test. The query-shape mirror between hook and invalidator is invariant-critical — a drift silently breaks invalidation — so a one-line shape assertion in `tests/runtime/dataviews-shared.test.mjs` would catch future drift.
- Same query shape is hand-built in two places (hook + invalidator). A tiny shared `buildCountQuery(field, value, baseQuery)` would remove the duplication. Cosmetic.

**Bottom line:** correct, complete, and addressed with care (good commit message, doc update, no scope creep). LGTM.

---

PR #39, opened 2026-05-14
Track E: PostsApp C2 hardening — i18n + view resync

dabowman:
## Summary

Lands the three C2 follow-ups for PostsApp documented in `docs/plans/track-e-postsapp-hardening.md`. **Blocker for Track F (entity-CRUD migration sweep).**

- **F1 — Render-time `__()` mapping for view-config labels.** `src/apps/posts/index.js` ships `FIELD_LABELS` + `ACTION_LABELS` constants; `buildFields` / `buildActions` consult `LABELS[id] ?? spec.label`. Translation tools see the `__()` literals at module load; plugin extension columns and actions fall through to the cascade-supplied label.
- **F2 — View-state resync on triple flip.** `useEffect` keyed `[ postType, variant ]` resets DataViews `view` to `{ ...VIEW_DEFAULTS, ...resolved.defaultView }` when the same hook instance hosts multiple triples. Effect deliberately doesn't depend on `viewConfig` so mid-session cascade re-resolves don't clobber in-session view edits.
- **F4 — `viewConfigFallback.js`.** Never existed. `inject_app_baselines` + `app.json#viewConfig` already cover every triple at boot; `useViewConfig` returns `config: doc ?? {}` for the empty-cascade edge. Verified zero references to a `viewConfigFallback` symbol anywhere in `src/` or `tests/`. Marked done as hygiene; no inbox row to move.

## Pattern documentation

Added to `CLAUDE.md` "Recurring patterns to enforce in review" so the next entity-CRUD migrations (taxonomy / users / comments / plugins / themes) copy the LABELS-table convention + view-resync `useEffect` verbatim. `src/apps/posts/app.md` rewrites the former "i18n regression" section as a "Translation recipe" with the precedence rule (LABELS wins for known ids; spec wins for unknown ids).

## Gates

This PR **must land before** any other entity-CRUD app migrates to the C2 view-config primitive — otherwise the i18n regression repeats per app.

## Test plan

- [x] `npm run lint:js` — clean
- [x] `npm run test:runtime` — 330 assertions pass
- [x] `npm run test:schema` — 79 assertions pass
- [ ] **Manual browser smoke under non-English locale** (e.g. `de_DE`) — confirm DataViews column headers + action labels render in that locale. Reviewer please verify before merge.
- [ ] Manual smoke: switch `config.postType` from `post` → `page` on the same mount, confirm clean view state with no stale filters / sort / perPage from the prior triple.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #93, opened 2026-05-28
Testing/0.1.0 testing

dabowman:
PR to collect feedback from testing and code review


dabowman:
## 🔍 Pre-Public-Testing Code Review

Whole-codebase audit for public-testing readiness. Scope: PHP (`includes/`), JS kernel (`src/runtime/`), apps (`src/apps/`), distribution, and `docs/` hygiene. **Desktop & single-pane engines excluded** (being extracted to plugins). Findings below were adversarially verified unless marked otherwise. Two candidates were investigated and **dropped as refuted**: the iframe-bridge `_parent`/`_top` "bypass" (the listener pins both `event.origin` and `event.source` to the bound same-origin frame) and the `/can/{capability}` "oracle" (it only probes the requesting user's own caps).

---

### 🔴 Tier 1 — Correctness / data-loss (fix before testers)

**1. `src/apps/simple-editor/index.js:256` — editor reuses one post's blocks for another → silent content corruption.**
`SimpleEditor`'s `hydrated` latch and `blocks` state never reset on `postId` change, and there's no `key={postId}`. `MountedApp` doesn't remount across same-route hash navs (`/posts/A/edit` → `/posts/B/edit` share one route pattern), and the outer `postId` `useState` initializer is also never re-synced. Opening post B in an already-mounted editor keeps post A's content; the 2 s autosave writes A's blocks against the pointed-at record. **Fix:** `key={postId}` on `<SimpleEditor>` (or reset `postId`/`hydrated`/`blocks` when `config.id` changes). Reachable in any shell routing `/posts/{id}/edit` to `core:simple-editor`.

**2. `src/apps/settings-general/index.js:29` — custom date/time format silently discarded on save.**
`dateFormatCustom`/`timeFormatCustom` seed from `useState(editedRecord?.date_format)`, but `editedRecord` is null on first render (Spinner returns before the record resolves) so both latch to `''` and never re-sync. On a site with a non-preset format, clicking a preset then re-clicking "Custom" without retyping writes `'Y-m-d'`, destroying the original. **Fix:** re-sync via `useEffect` on the resolved record.

**3. `src/apps/_shared/dataviews/createBulkConfirmModal.js:77` — double-click fires destructive mutations twice.**
The confirm button is only statically `disabled` via `isConfirmDisabled(targets)`; nothing guards re-entry while `Promise.allSettled(targets.map(mutate))` is in flight, and `closeModal()` runs after the await. A fast second click spawns a second delete/trash batch → duplicate DELETEs + spurious "N of N failed". Shared by all six entity-CRUD apps. **Fix:** busy flag / `loading` prop disabling the button on first click.

**4. `src/apps/taxonomy/index.js:159,259` — list stays stale after term delete/create/edit.**
`invalidateResolution('getEntityRecords', ['taxonomy', taxonomy])` omits the third `queryArgs` arg the live `useEntityRecords` (line 104) resolved under. `@wordpress/core-data` matches args by exact deep equality, not prefix — so the active query is never invalidated. Sibling apps (posts/users/comments) all pass the 3-element key; taxonomy is the outlier. **Fix:** pass `['taxonomy', taxonomy, queryArgs]`.

**5. `src/apps/plugins/index.js:263` — pagination broken above 50 plugins.**
The full filtered array is passed as `data` with `totalPages = ceil(data.length / perPage)`, but `data` is never sliced by `view.page`. DataViews is controlled and doesn't auto-slice, so page 1 shows all plugins and "Next" re-renders the same list. **Fix:** slice `data` by `view.page` (or paginate server-side).

---

### 🟠 Tier 2 — Security / information disclosure

**6. `wp-admin-shell.php:461` — full resolved `admin.json` shipped to every logged-in user; cap-pruning is client-side only.**
`wp_admin_shell_get_active_config()` returns the whole merged tree (every screen, each screen's `permissions` block, `legacy_path` maps, full menu IA) with no server-side prune — the resolver has zero `current_user_can` calls. A subscriber receives the complete admin IA in page source. Entity *data* is still REST-gated (structural/metadata leak, not record exposure). **Recommend:** prune screens/menu the user can't reach server-side before serialization.

**7. `src/apps/navigation/index.js:135` — role-gated nav items always visible client-side** *(accepted posture; pairs with #6)*.
`itemPassesPermissions` returns `true` for any `permissions:{roles:[...]}` item — there's no client-side role map. A `roles:['editor']` screen shows its label+href to a subscriber. Server 403s the route, but it compounds #6. Pruning server-side resolves both.

**8. `includes/class-wp-admin-shell-prefs-rest.php:63` — unbounded prefs merge → self-inflicted bloat/CPU.**
`POST /user-prefs` deep-merges arbitrary client JSON into usermeta with only a depth-10 cap — no byte/key bound (the file origin has a 1 MiB `MAX_BYTES` guard; this doesn't). The meta is re-read, merged, filtered, and JSON-encoded on every page load. Any logged-in user can POST a multi-MB object and degrade their own page loads + bloat the DB. **Fix:** add a size/key bound mirroring the file origin.

**9. `includes/cascade/class-wp-admin-shell-permissions.php:419 / :246` — empty-permission semantics diverge (latent)** *(defense-in-depth)*.
`resolve()` inflates an empty OR-set to admin-only; `user_passes()` treats the identical empty set as **allow-everyone**. Today the only behavioral caller pre-inflates via `resolve()`, so the allow-all branch isn't reached — but it's guarded by caller discipline, not by `user_passes`. Relatedly, a shrink-only consumer can't express "deny to nobody" (empty re-inflates to administrator). **Recommend:** make `user_passes` fail-closed on an empty set.

**10. `includes/cascade/class-wp-admin-shell-customizable.php:68` — `menu.*.permissions` absent from `DENY_PATTERNS`** *(defense-in-depth, visibility-only)*.
`screens.*.permissions` is deny-listed but `menu.*.permissions` isn't, and `bind_screens` won't overwrite a consumer-authored menu `permissions`. A consumer under a `customizable` ancestor could broaden nav-link visibility (real gates still hold → cosmetic). **Recommend:** add `menu.*.permissions` to the deny-list.

*Lower-confidence (not separately verified):* `class-wp-admin-shell-cache.php:88` `key_for()` omits engine `default-styles` + `wp_admin_shell_data_*` filter output from the cache signal → can serve stale resolved trees during testing. `class-wp-admin-shell-merge.php:110` `detect_key_field()` first-element-only fallback is latent fragility but not reachable in bundled configs (DENY_PATTERNS + keyed/assoc arrays block it).

---

### 🟡 Tier 3 — Distribution / packaging blockers

- **No `License`/`License URI` header** in `wp-admin-shell.php` (GPL only in README). Add `License: GPL-2.0-or-later` + `License URI:`; also missing `Author`/`Plugin URI`.
- **No `readme.txt`** — README.md isn't a substitute (Stable tag / "Tested up to" / sections; required for wp.org).
- **No `uninstall.php` / cleanup** — `wp_admin_shell_*` options + `wp_admin_shell_user_prefs` usermeta orphaned on uninstall; no clean reset between test runs.
- **No JS i18n** — no `wp_set_script_translations('wp-admin-shell', …)`, no `.pot`/`languages/`. Every `__()` string in the UI is untranslatable.
- **Gutenberg runtime-deactivation lockout** — dependency enforced only by an `admin_notices` warning; the hijack still boots and takes over `/wp-admin/`. Deactivating Gutenberg while the workspace is active renders the admin **blank**. Consider gating the hijack on a runtime check with classic fallback.
- **Version provenance** — header/`package.json` say `0.1.0`, but `CHANGELOG.md:76` references `1.0.0-beta.1`, the 0.1.0 entry sits under `[Unreleased]`, and there's no `WP_ADMIN_SHELL_VERSION` constant.
- **README leads with dev flow** — git-clone + `npm build` + placeholder `your-org` URL. Lead with the prebuilt-zip upload path.
- **Ships extracting engines** — `package.json` `files` bundles `core-desktop`/`core-single-pane` manifests + the four desktop `app.json`s; the PHP registry auto-registers them, so testers can select engines you intend to ship separately.

---

### 🟢 Tier 4 — `docs/` hygiene (clean to current-state only)

**Move to `docs/archive/`** (research/process/superseded): `feedback.md`, `feature-ideas.md`, `comms/`, `research/`, `desktop-engine-readiness.md`, `perf-baseline.md` (stale "v1", all pending), `wp-admin-screen-inventory.md` (overlaps `docs/screens/`), `wordpress-design-tokens-catalog.md` (hardcodes a local `/Users/...` path), completed `plans/`.

**Then update `CLAUDE.md`** — its "Before modifying code" list (items 9, 11) cites `feedback.md` and `desktop-engine-readiness.md` as canonical; those references dangle after the move.

**Keep** (canonical): `wp-admin-shell-design-spec.md`, `schema-sketch.md`, `schemas/`, `public/`, `screens/`, `code-map.md`, `dataview-config.md`, `engines-and-design-systems.md`, `core-default-engine.md`, `admin-json-api-validation.md`, `alpha-readiness.md`.

---

### ⚪ Tier 5 — Cleanup (post-testing OK)

- Chrome-hide CSS (`#adminmenuwrap`/`#wpadminbar`/…) is triplicated across `src/apps/editor/index.js:92`, `iframe-fallback`, and `desktop-iframe` — extract a `_shared/iframe/` helper.
- `src/apps/editor/index.js:101` — `catch(e){}` swallows non-cross-origin DOM errors silently; sibling code dev-warns.
- `src/runtime/registry/builtins.js:202` — kernel statically imports the desktop/single-pane engines + hardcodes the four desktop apps in `APP_LOADERS`; these become dangling references (build break) the moment those engines extract.

---

*Net: kernel DS-neutrality holds (no violations), the security model is mostly sound (REST is the real authority), and the cascade is well-tested. The Tier-1 data-loss bugs and Tier-3 distribution gaps are the real blockers for a clean public test.*

dabowman:
## 🔍 Code Review — Second Pass (new findings)

Follow-up to the existing Pre-Public-Testing review above. This pass re-swept the whole diff (4 parallel finder angles + WordPress-best-practice lens per the repo's `/wordpress-rest-api`, `/wordpress-dataviews`, `/wordpress-core-data` skills) **specifically for defects the first review did not already flag**. Each item below was verified against live source; I dropped several latent/refuted candidates (relative-`adminUrl` interceptor cluster — `adminUrl` is always `admin_url()`/absolute and the throw is caught → no live bug; media cross-page `invalidateResolution` staleness — inherent to client pagination, acceptable).

---

### 🔴 Correctness

**1. `src/apps/themes/index.js:269-271` — theme list pagination is fully suppressed.**
```js
totalItems: data.length,
totalPages: 1,
```
`data` is the entire unsliced record set and `totalPages` is hard-pinned to `1`, even though `VIEW_DEFAULTS.perPage = 50`. DataViews is controlled and won't auto-slice, so on an install with >50 installed themes every theme renders on one page and the pager is inert. Same class as the already-flagged plugins bug (#5) but stricter — there's no `view.page` slicing *and* no `totalPages` math at all. **Fix:** slice `data` by `view.page`/`view.perPage` and compute `totalPages`.

**2. `includes/cascade/class-wp-admin-shell-menu-items.php:469` — shallowest-wins dedupe drops a deduped node's *unique* children.**
```php
if ( isset( $screens[ $id ], $min_depths[ $id ] ) && $depth > $min_depths[ $id ] ) {
    continue; // ← skips BEFORE recursing into $item['items']
}
```
When a screen-bound id appears both shallow and deep, `drop_deeper_duplicates` `continue`s on the deeper occurrence *before* recursing into its `items`. If that deeper node is a container whose own children are screen-bound ids that appear **nowhere shallower**, those children are silently dropped along with the parent. **Trigger:** `menu.users` (screen `users`, depth 0) + `menu.admin.items.users` (screen `users`, depth 1) where the depth-1 node carries a unique `items.special-report` → `special-report` vanishes from the menu. **Fix:** recurse into `items` even when dropping the node itself, or only drop leaf duplicates.

**3. `src/apps/users/index.js` + `src/apps/comments/index.js` — display name / author name rendered without `decodeEntities`.**
Neither app imports `@wordpress/html-entities`, while `posts` and `taxonomy` both decode their name/title (per the CLAUDE.md "decodeEntities missing on titles/names" rule). A user named `Ben &amp;amp; Jerry` or a commenter `O&amp;#039;Brien` renders the raw entity string in the Name/Author column. (Comments' `content` is intentionally `dangerouslySetInnerHTML`'d and is fine — only the `author_name` text path is affected.) **Fix:** wrap the name/author display values in `decodeEntities()`.

---

### 🟠 Security / escape-hatch

**4. Non-admin lockout — workspace hijack floor is `read`, but the only forward escape (`?classic=1`) requires `manage_options`.**
`WP_Admin_Shell_Hijack::passes_base_gates()` (`class-wp-admin-shell-hijack.php:129`) takes over `/wp-admin/` for `current_user_can('read')` — i.e. every logged-in user, down to subscribers. But `WP_Admin_Shell_Classic_Mode` (`class-wp-admin-shell-classic-mode.php:76`) only honors `?classic=1` for `manage_options`. So an Author/Contributor/Subscriber who hits a workspace rendering bug **cannot reach classic wp-admin at all** — the toggle is silently ignored for them and no shipped app renders an escape control. (Relatedly, `README.md:96` promises admins "a Classic wp-admin toolbar button (session-scoped escape hatch)" that no bundled app — site-hub/user-menu/toolbar-actions — actually renders.) **Recommend:** lower the classic-toggle cap to match the hijack floor (`read`), and ship the documented escape control.

**5. `includes/cascade/class-wp-admin-shell-classic-menu-bridge.php:135` — the PR removed the self-reference skip-list while adding the very page it protected against.**
This diff deletes `'wp-admin-shell-settings'` / `'wp-admin-shell'` from `$CORE_SLUGS` (with the comment claiming "the shell's settings page is not surfaced as a third-party-plugin menu entry") — but `wp-admin-shell.php:319` now registers `add_options_page( …, 'wp-admin-shell-workspace', … )`, a non-core submenu child under `options-general.php` (a core parent). The bridge ingests non-core children of core parents, and `wp-admin-shell-workspace` matches neither the (deleted) skip entries nor any core slug, so the shell's own classic Settings page can surface as a synthesized `ingested-*` entry that links back into classic. The deletion's stated rationale is now false. **Fix:** add `'wp-admin-shell-workspace'` to the skip-list.

**6. `includes/class-wp-admin-shell-classic-mode.php:74-78` — `?classic=1` is a nonce-less state change (CSRF).** *(low severity)*
The toggle flips the `wp_admin_shell_classic` session cookie (which disables the workspace) on a cap-gated GET with no `wp_verify_nonce`. With the SameSite=Lax auth cookie, a top-level cross-site navigation to `/wp-admin/?classic=1` fires it for a logged-in admin. The state is reversible UI-only (no data write), so severity is low — but it's still a CSRF-able gate flip; the docblock waving it off as "idempotent" doesn't make it nonce-safe.

**7. `src/runtime/platform/iframeBridge.mjs:57-72` — the `_parent`/`_top` branch validates only origin, not the wp-admin path floor.** *(low severity)*
Every other classification path enforces `url.pathname.startsWith(base.pathname)`, but the `_parent`/`_top` branch checks only `expectedOrigin === targetOrigin` before returning `{ type: 'iframe', href: url }`. A same-origin-but-non-admin path (e.g. `/wp-content/uploads/x.html`) from the bound iframe is accepted into `iframe.src`. Source+origin pins limit *who* can send it (only the bound admin frame), so blast radius is small, but it widens the navigation sink beyond the admin tree, unlike the rest of the bridge.

---

### 🟡 Docs / distribution (deleted-shell dangling references)

This PR deletes `developer-admin` / `content-author` / `client-portal` / `canonical-demo` (only `wp-admin-default` / `desktop-demo` / `single-pane-demo` remain), leaving live references:

- **`README.md:85`** — the quickstart's **first command** `cp …/shells/developer-admin.json wp-content/admin.json` now fails (`No such file or directory`); a new user following the README verbatim ends up with no workspace.
- **`README.md:118-123`** — the starter-shells table still lists all four deleted shells, and the zip section claims it bundles "seven `shells/`" (now three).
- **`includes/class-wp-admin-shell-cli.php:59`** — `wp help admin-shell activate` example is `wp admin-shell activate content-author` → copy-paste yields `Shell not found`.

---

### ⚪ Lower-confidence (worth a glance)

- **`class-wp-admin-shell-classic-mode.php:96`** — `admin_bar_node()` shows the "Back to workspace" node on `! empty( $_COOKIE['wp_admin_shell_classic'] )`, but the hijack stands down only on `'1' === $_COOKIE[...]` (`hijack.php:121`). A forged/garbage truthy cookie (`=yes`) drifts the two predicates apart (cosmetic).
- **`includes/origins/class-wp-admin-shell-origin-file.php` `is_valid_partial()`** — rejects *scalar* known blocks but `is_array([])` is true, so a list-shaped block like `"screens": []` passes structural validation and flows into `merge_authoritative` against the assoc baseline. Trusted-origin author footgun; the `bad-screens-type.json` fixture covers the scalar case but not the empty-list/list case.

---

*Net: no kernel DS-neutrality violations, no new record-data exposure. The standout new items are the **non-admin classic lockout** (#4), the **menu-dedupe subtree loss** (#2), the **self-ingestion regression** (#5), and the **broken README onboarding path** (deleted shells). The lower-confidence pair is defense-in-depth.*

---

dabowman:
## 🔁 Re-review — fix commit verified

Re-reviewed the fix commit addressing both review passes (`c376e23` on `claude/pr-93-feedback-dpFkv` — not yet on the PR head `testing/0.1.0-testing`). Method: 3 parallel verification agents (PHP security / JS correctness / distribution) + a manual deep-read of the new server-side pruning. **For each fix I confirmed it addresses the finding and hunted for regressions the fix introduces.**

**Verdict: the fix commit is solid.** ~25 findings across both passes are correctly and completely addressed. One genuine new regression, plus a few minor/incomplete items below.

---

### 🔴 New regression (introduced by the fix)

**1. `wp-admin-shell.php:775` — server-side menu prune orphans *reachable* child screens when their screen-bound parent is pruned.**
The new `wp_admin_shell_prune_menu_for_user()` drops a node bound to a removed screen with a bare `continue` **before recursing into `$item['items']`**, so it discards the whole subtree — including children the user *can* still reach:
```php
if ( isset( $removed_screens[ (string) $id ] ) ) {
    continue; // ← drops $item['items'] too, even reachable ones
}
```
**Concrete case (wp-admin-default, any non-admin):** the `users` screen requires `list_users` (admin-only) → pruned → added to `$removed`. In the menu, the `profile` screen (requires only `read`) exists **only** nested at `menu.users.items.profile` — there's no top-level profile node. So `users` is dropped with its whole subtree, and **every non-admin (subscriber → editor) loses the Profile link from their menu**, even though the `profile` screen survives the screen-prune and is still URL-reachable. This is over-pruning, not a leak — but it's a visible parity regression (wp-admin always keeps Profile for non-admins).

Ironically the sibling `drop_deeper_duplicates()` fix in `class-wp-admin-shell-menu-items.php` *does* hoist a deduped node's surviving children — the prune function needs the same treatment: when a node is pruned but has reachable children, recurse and hoist them rather than dropping the subtree.

*(Uncertain, related: line 775 keys removal on screen id, so a manual non-screen-bound menu item whose id coincidentally equals a pruned screen id would also be dropped. Low likelihood — needs an id collision — flagging for awareness.)*

---

### 🟡 Incomplete / minor

**2. `docs/alpha-readiness.md:29` (and :5) — live smoke-test checklist still copies a deleted shell.** The README/CLI sweep fixed user-facing commands but missed this checklist: `Copy `shells/developer-admin.json` → `wp-content/admin.json`` → `No such file`. (The other ~18 `developer-admin`/`content-author`/etc. hits are in `docs/archive/`, `docs/screens/` archetype prose, and `app.md` — non-breaking.)

**3. `createBulkConfirmModal.js:83` — busy-flag double-click guard has no `try/finally` reset.** `isBusy` is set before the `await`, but `closeModal()` only runs on the success path after `onSettled?.()`/`onActionPerformed?.()`. If an app-supplied callback throws *after* the batch succeeded, the modal stays open with the confirm button stuck `disabled`. Edge case (callbacks are app-owned, unlikely to throw), but a `finally { setIsBusy(false) }` would harden it.

**4. Themes/plugins pagination fix works, but search/filter/sort remain unwired in controlled mode (pre-existing, not a regression).** Since DataViews is controlled (the app supplies `paginationInfo`), the app must derive `data` itself. `themes` `data` memo (`index.js:224`) only maps records — so typing in the search box, applying a status filter, or clicking a column header does nothing, and the new slice paginates an unsorted/unfiltered list. `plugins` applies search + status filter but not `view.sort`. Worth wiring (a `filterSortAndPaginate`-style derivation) so pagination is actually usable on installs with >50 themes/plugins.

---

### ✅ Verified correct & complete (no issues)

**Security:** server-side prune correctly evaluates OR-semantics via `resolve()` + `user_passes`, runs on a by-value copy (cache untouched), keeps `default-screen` (apart from finding #1) · `user_passes` empty-set fail-closed (real callers pre-inflate via `resolve()`, no regression) · prefs byte/key caps enforced before the merge · `menu.**.permissions` deny + correct `**`-glob matcher · `?classic=1` nonce verified against the matching action + cap floor lowered to `read` + admin-bar escape node + cookie predicate now `'1' ===` (matches hijack) · classic-menu-bridge skips `wp-admin-shell-workspace` · origin-file rejects list-shaped object blocks · Gutenberg gate stands the hijack down to classic instead of rendering blank.

**Correctness:** simple-editor `key={postId}` + resync (no cross-post bleed/autosave) · settings-general one-shot format resync reads `record` and only touches shadow state (no clobber) · taxonomy uses the full 3-element invalidation key · users/comments `decodeEntities` on name/author · shared `chromeHide.mjs` preserves all selectors + dev-warns on non-SecurityError · iframeBridge `_parent`/`_top` now enforces the wp-admin path floor · menu-items dedupe hoist re-parents unique children without reintroducing duplicates.

**Distribution:** `uninstall.php` (WP_UNINSTALL guard, all 8 options + `LIKE` sweep with `esc_like`, prefs usermeta via `delete_metadata(..., true)`, transients, multisite loop) · `readme.txt` (Stable tag / Requires / Tested up to / License / sections) · header License/Author/Plugin URI + `WP_ADMIN_SHELL_VERSION` · i18n `wp_set_script_translations` + `.pot` (handle/domain/path correct) · version `0.1.0` consistent across header / constant / package.json / readme / .pot · README/CLI deleted-shell refs cleaned.

*(Minor: `uninstall.php` deletes the global user-prefs usermeta once per blog in the multisite loop — harmless/idempotent, just redundant queries on large networks.)*

---

*Net: ship-worthy once the **Profile/menu over-prune (#1)** is fixed — that's the only behavior regression. #2–#4 are polish.*

---

dabowman:
## ✅ Re-review (round 3) — all clear

Reviewed `50a0dc6` ("fix menu over-prune regression + polish"), now on the PR head `testing/0.1.0-testing`. **All four items from the previous re-review are correctly fixed, and the commit introduces no new regressions.**

- **🔴 Menu over-prune (the one real regression) — FIXED.** `wp_admin_shell_prune_menu_for_user()` now recurses into `items` *first*, then on a dropped node hoists the surviving children up to the dropped node's level (with an `! isset( $out[ $child_id ] )` collision guard), instead of `continue`-ing past the whole subtree. Mirrors `drop_deeper_duplicates()`. The Profile case resolves: a subscriber loses `users` but keeps a hoisted top-level `profile`. Backed by a new `run-cap-gating-smoke.php` regression that asserts exactly that (subscriber: no `users` screen/node, keeps `profile` screen + hoisted menu node; admin: full tree intact). PHP lints clean.
- **createBulkConfirmModal — FIXED.** Confirm handler wrapped in `try { … closeModal() } finally { setIsBusy( false ) }`, so a throwing app `onSettled`/`onActionPerformed` can no longer leave the button stuck disabled. (`Promise.allSettled` itself never rejects, as the comment notes.)
- **Themes search/filter/sort — FIXED properly.** Switched to the package's own `filterSortAndPaginate( data, view, fields )` from `@wordpress/dataviews/wp` (correct private subpath) returning `{ data, paginationInfo }` — search box, status filter, column sort, and pager all work now, not just pagination.
- **Plugins sort — FIXED.** Adds `view.sort` with `localeCompare(…, { numeric: true })` (version segments collate correctly) before the page slice; keeps the bespoke search that also covers the description rendered in the name cell. `totalPages` still derives from the filtered length. Reasonable that it didn't adopt `filterSortAndPaginate` (global search wouldn't cover the description).
- **`docs/alpha-readiness.md` — FIXED.** Smoke checklist now copies `single-pane-demo.json` (an existing shell) instead of the deleted `developer-admin.json`.

*Couldn't run `npm run lint:js` (deps not installed in the review container), but the changes follow repo conventions and PHP lints clean.*

**Net: the PR is clean from a code-review standpoint.** Both review passes and both re-reviews are resolved; no outstanding correctness or security findings. The remaining pre-existing notes (the earlier review's defense-in-depth items already accepted as posture, manual smoke/perf/a11y passes) are out of code-review scope.

---

dabowman:
when a page is updated in the iframed edit app the view page link in the toast notification doesn't work.

Issue #143, opened 2026-05-29
[upstream] Writable theme status (activate) + DELETE on the themes controller


dabowman:
WP_REST_Themes_Controller is read-only; switch_theme() and theme deletion have no REST route. Add a writable status and a DELETE route. Source: docs/parity/roadmap.md (group B, P1, item 5); detail in docs/parity/themes.md. Parity audit, PR #96.

PR #178, opened 2026-05-29
fix(themes): register activate-theme REST route; drop nonce-less fallback

### LINKED ISSUE #97: Fix theme Activate: it POSTs to an unregistered endpoint and silently fails
Activate POSTs to an endpoint that is not registered, then falls back to a nonce-less wp-admin link that silently fails; theme activation is non-functional on a clean install. Surface an error Notice on failure. Source: docs/parity/roadmap.md (group A, P1, item 1); detail in docs/parity/themes.md. Parity audit, PR #96.

dabowman:
**Closes #97.**

Theme **Activate** was POSTing to `/wp-admin-shell/v1/activate-theme` — a route that was never registered — then falling back to a nonce-less `themes.php?action=activate` link that silently failed. Activation was non-functional on a clean install. (Core has no writable themes REST; that's upstream #143. This is the shell-side workaround.)

- New `WP_Admin_Shell_Themes_REST` registers `POST /wp-admin-shell/v1/activate-theme` (mirrors the can/prefs controller pattern), gated on `current_user_can('switch_themes')`, `stylesheet` sanitized; validates via `wp_get_theme()` (400 empty / 404 missing / 400 broken) then `switch_theme()`.
- ThemesApp now shows an **error Notice** on failure instead of the nonce-less redirect; success path (snackbar + `root/theme` invalidation) unchanged.

**Tests:** `lint:js` ✅ · `test:runtime` (33 files) ✅ · `php -l` ✅ · PHP cap tests **UNVERIFIED** (wp-env not running). Follow-up: a `switch_themes` permission-floor test (subscriber 403 / 404 / 400) would be worth adding.

### DIFF
--- a/includes/class-wp-admin-shell-themes-rest.php
+++ b/includes/class-wp-admin-shell-themes-rest.php
@@ -60,8 +60,10 @@ class WP_Admin_Shell_Themes_REST {
 	 * @return WP_REST_Response|WP_Error
 	 */
 	public static function activate( $request ) {
+		// REST sanitizes `stylesheet` to a string before dispatch; `required`
+		// asserts presence, not non-emptiness, so an empty value reaches here.
 		$stylesheet = $request->get_param( 'stylesheet' );
-		if ( ! is_string( $stylesheet ) || '' === $stylesheet ) {
+		if ( '' === $stylesheet ) {
 			return new WP_Error(
 				'rest_invalid_param',
 				__( 'A theme stylesheet is required.', 'wp-admin-shell' ),
@@ -78,12 +80,36 @@ class WP_Admin_Shell_Themes_REST {
 			);
 		}
 
+		// Mirror core themes.php: on multisite a site admin with `switch_themes`
+		// must not activate a theme the super-admin network/site-disabled.
+		// `is_allowed()` returns true on single-site, so this is multisite-only.
+		if ( ! $theme->is_allowed() ) {
+			return new WP_Error(
+				'rest_theme_not_allowed',
+				__( 'This theme is not allowed on this site.', 'wp-admin-shell' ),
+				array( 'status' => 403 )
+			);
+		}
+
 		$errors = $theme->errors();
 		if ( $errors instanceof WP_Error ) {
 			return new WP_Error(
 				'rest_theme_broken',
 				/* translators: %s: theme error message. */
 				sprintf( __( 'The theme cannot be activated: %s', 'wp-admin-shell' ), $errors->get_error_message() ),
+				array( 'status' => 400 )
+			);
+		}
+
+		// `WP_Theme::errors()` reports only missing/broken files — never
+		// RequiresWP/RequiresPHP. Without this pre-check an intact-but-
+		// incompatible theme reaches switch_theme(), which wp_die()s mid-REST
+		// (a 500 carrying literal <strong> markup). Validate up front instead.
+		$requirements = validate_theme_requirements( $theme->get_stylesheet() );
+		if ( is_wp_error( $requirements ) ) {
+			return new WP_Error(
+				'rest_theme_requirements',
+				wp_strip_all_tags( $requirements->get_error_message() ),
 				array( 'status' => 400 )
 			);
 		}
--- a/includes/class-wp-admin-shell-themes-rest.php
+++ b/includes/class-wp-admin-shell-themes-rest.php
@@ -1,0 +1,103 @@
+<?php
+/**
+ * /wp-admin-shell/v1/activate-theme — switch the active theme.
+ *
+ * Shell-side workaround: WordPress core ships no writable themes REST
+ * endpoint (upstream parity ticket #143), so `core:themes` has no
+ * canonical `/wp/v2/themes` mutation to call. This endpoint is a thin
+ * transport over `switch_theme()`, gated on `switch_themes`.
+ *
+ * The previous client fallback navigated to a nonce-less
+ * `themes.php?action=activate&...` link, which silently failed — so theme
+ * activation was non-functional on a clean install. This route replaces
+ * that fallback. apiFetch's `wp-api-fetch` middleware sends the REST
+ * nonce automatically.
+ *
+ * @package WP_Admin_Shell
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+class WP_Admin_Shell_Themes_REST {
+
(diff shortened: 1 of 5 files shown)


includes/class-wp-admin-shell-themes-rest.php
```diff
@@ -0,0 +1,103 @@
...
+		if ( $errors instanceof WP_Error ) {
+			return new WP_Error(
+				'rest_theme_broken',
+				/* translators: %s: theme error message. */
+				sprintf( __( 'The theme cannot be activated: %s', 'wp-admin-shell' ), $errors->get_error_message() ),
+				array( 'status' => 400 )
+			);
+		}
+
+		switch_theme( $theme->get_stylesheet() );
```
dabowman, line 117: **Missing `validate_theme_requirements()` pre-check (correctness/parity).**

`switch_theme()` itself calls `validate_theme_requirements()` and `wp_die()`s on a WP/PHP-incompatible theme (`wp-includes/theme.php`). `WP_Theme::errors()` reports only missing/broken files — never `RequiresWP`/`RequiresPHP` — so an intact-but-incompatible theme passes both guards above and reaches `switch_theme()`, which `wp_die()`s mid-REST. The JSON die handler returns a **500** (semantically should be 400) whose message carries literal `<strong>` tags that the snackbar renders verbatim.

The site is *not* switched (it dies before switching), so no white-screen — but core's clean validation path is bypassed. Fix: call `validate_theme_requirements( $theme->get_stylesheet() )` before `switch_theme()` and return a 400 `WP_Error` when it's a `WP_Error`.

includes/class-wp-admin-shell-themes-rest.php
```diff
@@ -0,0 +1,103 @@
...
+		if ( ! is_string( $stylesheet ) || '' === $stylesheet ) {
+			return new WP_Error(
+				'rest_invalid_param',
+				__( 'A theme stylesheet is required.', 'wp-admin-shell' ),
+				array( 'status' => 400 )
+			);
+		}
+
+		$theme = wp_get_theme( $stylesheet );
+		if ( ! $theme->exists() ) {
```
dabowman, line 75: **Missing multisite `$theme->is_allowed()` gate (security/parity).**

Core's `themes.php` gates activation on `! $theme->exists() || ! $theme->is_allowed()` (line 25). On multisite, a site admin who has `switch_themes` can POST a stylesheet that exists on disk but was network/site-disabled by the super-admin, and it activates — a policy bypass not reachable through classic wp-admin. `is_allowed()` returns `true` on single-site, so this is multisite-only.

Fix: add `! $theme->is_allowed()` to the validation here (403).

dabowman (commented):
## Automated code review (`/code-review max`)

Reviewed the new endpoint against WordPress core's own `wp-admin/themes.php` activation handler. The core premise is sound — replacing the broken nonce-less `window.location.href` fallback with a real REST endpoint + error notice is the right move, and the capability string (`switch_themes`), nonce handling (apiFetch middleware), and `useCallback` deps are all correct.

Two correctness/parity gaps and one UX issue are worth addressing before merge. Inline comments below. One finding falls outside the changed hunk so it's here:

### `src/apps/themes/index.js` (~line 211) — details-modal Activate signals false success on failure
The modal's Activate button does `await activate( item ); closeModal();`. `activate()` catches its own error and never rethrows, so the `await` always resolves and `closeModal()` runs even when activation failed — the modal vanishes as if it succeeded, with only the separate error snackbar hinting otherwise (no in-modal error, no retry affordance). The row-action path is fine. Consider having `activate()` surface its outcome so the modal can keep itself open on failure (cf. `createBulkConfirmModal`, which only `closeModal()`s on the success path).

---

includes/class-wp-admin-shell-themes-rest.php
```diff
@@ -0,0 +1,103 @@
...
+
+	/**
+	 * Validate the requested theme and switch to it.
+	 *
+	 * @param WP_REST_Request $request Request.
+	 * @return WP_REST_Response|WP_Error
+	 */
+	public static function activate( $request ) {
+		$stylesheet = $request->get_param( 'stylesheet' );
+		if ( ! is_string( $stylesheet ) || '' === $stylesheet ) {
```
dabowman: **Minor — dead branch.** `stylesheet` is declared `'type' => 'string'` + `'sanitize_callback' => 'sanitize_text_field'`, and REST runs `sanitize_params()` (always yields a string) before dispatch, so `! is_string( $stylesheet )` is unreachable. The empty-string check *is* load-bearing (`'required' => true` asserts presence, not non-emptiness — `stylesheet=''` reaches here). Can simplify to `if ( '' === $stylesheet )`.

includes/class-wp-admin-shell-themes-rest.php
```diff
@@ -0,0 +1,103 @@
...
+ * nonce automatically.
+ *
+ * @package WP_Admin_Shell
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+class WP_Admin_Shell_Themes_REST {
+
+	const NAMESPACE = 'wp-admin-shell/v1';
```
dabowman, line 23: **Nit (pre-existing pattern, not introduced here).** This is the 5th `*-rest.php` controller to hardcode `'wp-admin-shell/v1'`, and the siblings are split between `const NAMESPACE` and `const REST_NAMESPACE`. A future namespace bump touches every file — worth a one-time consolidation into a single shared `WP_ADMIN_SHELL_REST_NAMESPACE` define.

dabowman: Agreed this is worth doing, but since it's a pre-existing cross-file pattern (5 controllers, split between `NAMESPACE`/`REST_NAMESPACE`) and not introduced here, I'm leaving it out of this PR to keep the diff focused on the activate-theme fix. Tracking it as a separate consolidation cleanup. The other three threads (requirements pre-check, multisite `is_allowed()` gate, dead `is_string` branch) are fixed in 8d5d9f9, along with the details-modal false-success issue from the review summary.

---

dabowman: Correction: the fixes landed in `83dff44`, not `8d5d9f9` (typo above).

---

dabowman:
Superseded by #188 — the identical 5-file change was rebuilt cleanly on current `main` (this branch was 5 commits stale off its base) and merged in `5780ea6`, closing #97. Closing this in favor of that.

---

PR #179, opened 2026-05-29
chore(_shared): wire DataForm validation + drop dead isDestructive

### LINKED ISSUE #130: Wire useFormValidity into EntityDataForm
Pass validity and gate Save on it; fields already declare isValid rules that are currently inert across all form apps. Source: docs/parity/roadmap.md (group A, P3, item 34); detail in docs/parity/dataviews-dataforms-limitations.md. Parity audit, PR #96.

### LINKED ISSUE #135: Drop the dead isDestructive mapping in _shared/dataviews/buildActions.js
DataViews ignores it; cross-cutting across all six list apps. Source: docs/parity/roadmap.md (group A, P3, item 39); detail in docs/parity/dataviews-dataforms-limitations.md. Parity audit, PR #96.

dabowman:
**Closes #130, Closes #135.** Two `_shared` foundation cleanups (land before fanning out wave:2, which builds on this scaffolding).

**#130 — wire `useFormValidity`** (`EntityDataForm.js`): fields declared `isValid` rules (e.g. taxonomy `name: { required: true }`) that were inert — `EntityDataForm` never passed `validity` and gated Save only on `hasEdits`. Now calls `useFormValidity(item, fields, form)` from `@wordpress/dataviews/wp` (v14.0.0, the package's own canonical validation pattern), passes `validity` to `DataForm`, and gates Save on `! isValid`. Hook runs before the null-guard so React order is stable; returns `isValid: true` for an empty/loading record so Save isn't spuriously disabled during load.

**#135 — drop dead `isDestructive`** (`buildActions.js`): the `@wordpress/dataviews` `Action` type has no `isDestructive` field, so the compiled assignment was inert metadata. Removed it + the JSDoc mention. Grep-verified nothing reads it (the other `isDestructive` hits are legitimate `@wordpress/components` Button props in modals).

**Tests:** `test:runtime` ✅ (all suites, 0 failures) · `lint:js` ✅. Shared test file needed no change (it pins only the pure helpers).

### DIFF
--- a/src/apps/_shared/dataviews/buildActions.js
+++ b/src/apps/_shared/dataviews/buildActions.js
@@ -5,9 +5,9 @@ import { compileEligibility } from './compileEligibility.mjs';
  * Shared DataViews action compiler for the entity-CRUD apps.
  *
  * Every app compiles the same action shape — `{ id, label, isPrimary,
- * isDestructive, supportsBulk, icon, isEligible }` — then attaches EITHER a
- * `RenderModal` (confirmation / detail modals) OR a `callback` (fire-and-go
- * actions). The differences are entirely in the data passed in:
+ * supportsBulk, icon, isEligible }` — then attaches EITHER a `RenderModal`
+ * (confirmation / detail modals) OR a `callback` (fire-and-go actions). The
+ * differences are entirely in the data passed in:
  *
  * - `labels`               — ACTION_LABELS table (id → translated label).
  * - `callbacks`            — id → `(items) => void` for simple actions.
@@ -39,7 +39,6 @@ export function buildActions(
 				id: spec.id,
 				label: labels[ spec.id ] ?? spec.label,
 				isPrimary: !! spec.isPrimary,
-				isDestructive: !! spec.isDestructive,
 				supportsBulk: !! spec.supportsBulk,
 				icon: spec.icon ? resolveIcon( spec.icon ) : undefined,
 				isEligible:
--- a/src/apps/_shared/forms/EntityDataForm.js
+++ b/src/apps/_shared/forms/EntityDataForm.js
@@ -1,5 +1,5 @@
 import '../app.css';
-import { DataForm } from '@wordpress/dataviews/wp';
+import { DataForm, useFormValidity } from '@wordpress/dataviews/wp';
 import { useEntityRecord } from '@wordpress/core-data';
 import { Button, Stack, Text } from '@wordpress/ui';
 import { Spinner } from '@wordpress/components';
@@ -43,6 +43,15 @@ export function EntityDataForm( {
 		useEntityRecord( ...entity );
 	const handleSave = useEntitySave( save, messages );
 
+	// Wire `isValid` field rules into live validation. Must run before the
+	// null-guard early return so the hook order stays stable across renders;
+	// `useFormValidity` reports `isValid: true` for an empty/loading record.
+	const { validity, isValid } = useFormValidity(
+		editedRecord ?? {},
+		fields,
+		form
+	);
+
 	if ( ! record ) {
 		return (
 			<div className="wp-admin-shell-app__center">
@@ -67,6 +76,7 @@ export function EntityDataForm( {
 					data={ editedRecord }
 					fields={ fields }
 					form={ form }
+					validity={ validity }
 					onChange={ edit }
 				/>
 				{ children }
@@ -75,7 +85,7 @@ export function EntityDataForm( {
 						tone="brand"
 						variant="solid"
 						onClick={ handleSave }
-						disabled={ ! hasEdits || isSaving }
+						disabled={ ! hasEdits || ! isValid || isSaving }
 						loading={ isSaving }
 					>
 						{ saveLabel || __( 'Save Changes', 'wp-admin-shell' ) }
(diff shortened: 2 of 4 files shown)


dabowman (commented):
## Code review (max effort)

Reviewed in an isolated worktree; analyzed the `@wordpress/dataviews@14.0.0` internals (`useFormValidity`, `getIsValid`, the `text`/`integer`/`elements` validators) and all three `EntityDataForm` callers, cross-checked by two independent verification agents.

**`buildActions.js` (#135) — ✅ safe.** DataViews never reads `action.isDestructive` off a compiled action (the `isDestructive` hits in the bundle are internal `Text`/`Button` style props); no app reads it, no test asserts it, and the schema/fixtures still legally carry it. Fully inert removal.

**`EntityDataForm.js` (#130) — ⚠️ behavior change with a Save-lock risk.** The import resolves fine (`useFormValidity` is exported from the `/wp` build, so no crash). But wiring `useFormValidity` into the *shared* form unconditionally enables DataViews' **implicit element-membership validation**. None of the callers declare `isValid` rules, yet `getIsValid` auto-enables an `elements` validator for any field carrying an `elements` array — and those option lists are capped at `per_page: 100` and lazy-loaded, while `getFormFieldsToValidate` ignores `isVisible`.

### 1. `page_on_front` / `page_for_posts` can disable Save — `EntityDataForm.js:88` + `settings-reading/index.js:47,81,93`
`pageOptions` is built from `per_page:100, status:'publish'` and the fields are `type:'text'` with `elements:pageOptions`, so `isValidElements` is auto-enabled.
- **Durable:** if the static front-page id isn't in the first 100 published pages (site has >100 pages, or it's a draft/private page) → `isValid=false` → Save disabled even when editing an unrelated field like `posts_per_page`.
- **Transient (every load):** while the pages query resolves, `pageOptions` is just `[{value:'0'}]`, so any non-zero `page_on_front` is invalid and Save is dead until pages load.

### 2. `default_category` can disable Save — `settings-writing/index.js:27,44`
Same class via pagination: `categoryOptions` is capped at `per_page:100`. On a site with >100 categories where the default sits beyond the fetched page (or `default_category` is 0/unset → `'0'`/`''` not in the option list, which has no sentinel), the whole Writing panel becomes un-saveable.

### 3. Hidden fields are still validated → **invisible** Save-lock — `EntityDataForm.js:49`
`getFormFieldsToValidate` never consults `isVisible`. `page_on_front`/`page_for_posts` carry `isVisible: showsStaticPage` (`settings-reading/index.js:86,98`) and are hidden when `show_on_front='posts'`. A site that switched from a static homepage back to "latest posts" keeps a stale non-zero `page_on_front`; if that id is outside the first-100 pages, Save is disabled while the offending select isn't even rendered — undiagnosable. Generalizes to any future hidden field with an `elements` list added to these shared forms.

### 4. (minor) Comment mislabels the mechanism — `EntityDataForm.js:46-48`
No caller declares `isValid` rules; what's actually wired is the type-default `elements` validator. And `isValid: true` for an empty record holds only on the first synchronous render — after the post-paint effect runs `validate()` on `{}`, `show_on_front` (getValue→`undefined`, ∉ `['posts','page']`) is marked invalid. It's harmless only because the spinner hides the button during load, not because the hook returns `true`.

### 5. (minor) `editedRecord ?? {}` is a dead guard — `EntityDataForm.js:50`
`getEditedEntityRecord` returns `EMPTY_OBJECT` or `{...record,...edits}` — never null/undefined — so the `?? {}` fallback can't fire.

---

**Suggested direction:** opt the affected selects out of implicit validation (`isValid: { elements: false }`), or always inject the current stored value into the options list, or gate Save only on author-declared validity. Per-field opt-out is safer than suppressing validation globally now that wave:2 will lean on it. Findings 1–2 are the ones to resolve before merge; 3 shares their root but is the nastiest in character.

<sub>🤖 automated review via /code-review</sub>

dabowman:
Addressed in e2945f6. Verified the review against the `@wordpress/dataviews@14.0.0` source before touching anything — the findings hold:

- `getIsValid` (`field-types/utils/get-is-valid.mjs`) auto-enables the `elements` validator whenever a field carries an `elements`/`getElements` and doesn't declare `isValid.elements` — the source comment literally reads *"elements is enabled unless the field opts-out."*
- `isValidElements` fails membership when the stored value isn't in `elements[].value`, and `getFormFieldsToValidate` never consults `isVisible` (so hidden fields validate too — confirming #3).
- `Rules.elements` is typed `boolean`, so `isValid: { elements: false }` is the canonical opt-out.

Fixes:

- **#1 / #3** — `page_on_front` + `page_for_posts` (settings-reading) → `isValid: { elements: false }`. Kills both the >100-pages / draft-page lock and the hidden-field lock.
- **#2** — `default_category` (settings-writing) → same opt-out (pagination + unset-`''` case).
- **#4** — rewrote the `EntityDataForm` comment to describe the actual mechanism (auto-enabled `elements` check + the spinner gate masking any transient `isValid:false`), not author-declared rules.
- **#5** — dropped the dead `editedRecord ?? {}` guard (`getEditedEntityRecord` never returns null).

Went with **per-field opt-out** rather than global suppression, per your suggested direction — keeps the validation infra live for author-declared rules in wave:2. Static-element fields (`show_on_front`, `rss_use_excerpt`, `default_post_format`) stay validated: their values come from closed sets, so membership is correct and can't lock Save. `settings-discussion` is all booleans (unaffected); `profile`/taxonomy use `DataForm` directly, not `EntityDataForm`.

`lint:js` ✅ · `test:runtime` ✅ (all suites, 0 failures).

---

PR #180, opened 2026-05-29
fix(dataviews): persist per-screen view to user prefs (Screen Options parity)

### LINKED ISSUE #131: Persist DataViews view (column-hide/sort/perPage) to user prefs
Via the existing /prefs REST so Screen-Options-equivalent state survives navigation/reload. The biggest Screen Options gap. Source: docs/parity/roadmap.md (group A, P3, item 35); detail in docs/parity/dataviews-dataforms-limitations.md. Parity audit, PR #96.

dabowman:
**Closes #131.**

DataViews view state (column visibility / sort / per-page / layout — the wp-admin Screen Options equivalent) was lost on navigation/reload because `useEntityDataView`'s resync effect re-seeded from `defaultView` every time. Now persisted per-screen to user prefs and rehydrated.

- New pure helper `dataViewPrefs.mjs` (`pickDurableView` / `readSavedView` / `buildSavePatch` / `applySavedView` + `DURABLE_AXES`).
- Persists **durable** axes only (type/fields/sort/perPage/layout/title-media-description) — **excludes** transient `search`/`filters`/`page` so a deep-linked filter isn't clobbered. Matches what Screen Options remembers.
- Reuses the existing `/wp-admin-shell/v1/user-prefs` REST surface (same one `core:appearance` uses) via `apiFetch`; storage under a `dataViews` key sub-keyed by `screenId`. Writes debounced 600ms, no-op/transient-only changes skipped.
- Reconciliation: synchronous seed renders first (no flash); a baseline ref prevents the rehydrated value being written straight back; genuinely different screens still re-seed, but the saved view wins over `defaultView` for the same screen. Title-dedup + selection reset unchanged; no call-site changes.

**Tests:** `test:runtime` ✅ (dataviews-shared: 47 passed, +17 new) · `lint:js` ✅ · **REST smoke against wp-env VERIFIED** — blob stores/reads through prefs meta, and the resolver drops the `dataViews` key from resolved config (no cascade pollution).

### DIFF
--- a/src/apps/_shared/dataviews/dataViewPrefs.mjs
+++ b/src/apps/_shared/dataviews/dataViewPrefs.mjs
@@ -1,0 +1,118 @@
+/**
+ * Pure helpers for persisting a DataViews `view` to user preferences —
+ * the Screen-Options-equivalent (column visibility, sort, per-page, layout).
+ *
+ * Storage lives in the shell's `wp_admin_shell_user_prefs` user-meta, written
+ * through `POST /wp-admin-shell/v1/user-prefs` (partial deep-merge, `null`
+ * tombstones a key). The blob lives under a top-level `dataViews` key, sub-
+ * keyed by `screenId`:
+ *
+ *   { "dataViews": { "posts": { "fields": [...], "sort": {...}, ... } } }
+ *
+ * The cascade resolver's consumer-origin `customizable` walker only copies the
+ * recognized admin.json blocks (`settings` / `styles` / the v3 top-level
+ * blocks) into the merged doc, so a `dataViews` top-level key is **silently
+ * dropped from the resolved config** — no pollution — while remaining stored in
+ * the meta and readable via the `/user-prefs` GET endpoint (same pattern the
+ * `shell` / `default-route` prefs already use). These helpers are side-effect
+ * free so `tests/runtime/dataviews-shared.test.mjs` can import them directly.
+ */
+
+/** Top-level user-prefs key the view blobs live under. */
+export const PREFS_KEY = 'dataViews';
+
+/**
+ * View axes that persist (the Screen-Options-equivalent durable state). These
+ * survive navigation + reload. `search` / `filters` / `page` are deliberately
+ * excluded — they're transient query state, not remembered preferences (this
+ * matches what wp-admin Screen Options remembers: hidden columns + per-page,
+ * not the current search/filter/page).
+ */
+export const DURABLE_AXES = [
+	'type',
+	'fields',
+	'sort',
+	'perPage',
+	'layout',
+	'titleField',
+	'mediaField',
+	'descriptionField',
+	'showTitle',
+	'showMedia',
+	'showDescription',
+];
+
+/**
+ * Project a full `view` down to only the durable axes worth persisting.
+ *
+ * @param {Object} view A DataViews `view` object.
+ * @return {Object} A new object containing only present durable axes.
+ */
+export function pickDurableView( view ) {
+	const out = {};
+	if ( ! view || typeof view !== 'object' ) {
+		return out;
+	}
+	for ( const axis of DURABLE_AXES ) {
+		if ( view[ axis ] !== undefined ) {
+			out[ axis ] = view[ axis ];
+		}
+	}
+	return out;
+}
+
+/**
+ * Read the saved durable view for a screen out of a full prefs blob.
+ *
+ * @param {Object} prefs    The full user-prefs object (from GET /user-prefs).
+ * @param {string} screenId The active screen id.
+ * @return {Object|null} The saved durable view, or null when none is stored.
+ */
+export function readSavedView( prefs, screenId ) {
+	if ( ! prefs || typeof prefs !== 'object' || ! screenId ) {
+		return null;
+	}
+	const bucket = prefs[ PREFS_KEY ];
+	if ( ! bucket || typeof bucket !== 'object' ) {
+		return null;
+	}
+	const saved = bucket[ screenId ];
+	if ( ! saved || typeof saved !== 'object' ) {
+		return null;
+	}
+	return saved;
+}
+
+/**
+ * Build the partial POST body that persists a screen's durable view.
+ *
+ * The server deep-merges this onto the existing prefs, so only the touched
+ * screen's bucket is written; sibling screens' saved views are untouched.
+ *
+ * @param {string} screenId The active screen id.
+ * @param {Object} view     The full or durable `view` to persist.
+ * @return {Object|null} The POST body, or null when there's nothing to write.
+ */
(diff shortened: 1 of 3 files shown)


dabowman:
@claude review this pr

dabowman:
## Code review — max-effort pass 🤖

Reviewed via multi-angle finders + verification + a gap sweep against a worktree at the PR head. The **headline bug was reproduced empirically** by executing the server `deep_merge` directly, and the test suite was run (47 pass — but they don't exercise the path that's broken).

**Verdict:** the feature doesn't work as intended and *corrupts* stored state on the second column-hide, plus a cluster of races in the new async plumbing and a site-wide cache regression. The green tests miss all of it because they only cover the pure JS helpers, never the PHP transport.

### Findings (ranked)

**1. 🔴 `deep_merge` index-merges JSON arrays → hiding a column corrupts/ignores the saved `fields` list** · `includes/class-wp-admin-shell-prefs-rest.php:162`
This PR is the first to persist a list-valued axis through a `deep_merge` previously used only for scalars (`shell`, `default-route`). It recurses into arrays by numeric index and never deletes trailing elements. Reproduced by running the function:
- hide a column (`['title','author','date']` ← `['title','date']`) → stored **`["title","date","date"]`** (duplicate; hidden col survives)
- hide down to one (`['title']`) → stored **`["title","author","date"]`** (hide *completely ignored*)
- first-ever save and *growing* the list work — masking the bug in casual testing.

Defeats the stated purpose ("remember hidden columns"). Right fix is server-side: replace `array_is_list($over)` arrays wholesale instead of recursing (the client already sends the full authoritative list).

**2. 🔴 Appearance "Reset" wipes every saved DataViews view** · `prefs-rest.php:101` + `src/apps/appearance/index.js:134`
`reset()` → `DELETE /user-prefs` → `delete_user_meta()` nukes the **entire** blob. The new `dataViews` bucket shares that meta key, so "Reset to shell defaults" (intending to revert density/accent) silently discards all column/sort/perPage prefs across every list screen.

**3. 🔴 Mount `useEffect([])` captures a stale `screenId`; a late GET overlays the wrong screen's saved view** · `src/apps/_shared/dataviews/useEntityDataView.js:91`
The hook instance survives screen flips (no `key` on `MountedApp`; same component reconciled — which is *why* the resync effect exists). Navigate `/posts → /pages` before the GET resolves and the `.then` reads `readSavedView(prefs, 'posts')` (mount-time capture) and overlays posts' columns onto the pages view. `cancelled` only guards unmount, not a flip.

**4. 🔴 Late mount GET clobbers the user's in-flight edit** · `useEntityDataView.js:98`
No "user-touched" guard. Toggle a column while the GET is in flight → `setRawView(current => applySavedView(current, saved))` overwrites the change with the saved view (lost update).

**5. 🔴 Shared `persistTimer` ref drops a pending write across a screen flip** · `useEntityDataView.js:152`
Edit a column on screen A (schedules a 600ms write), navigate to B, edit a durable axis on B within the window → the persist effect's `clearTimeout(persistTimer.current)` cancels A's still-pending write. Single shared slot conflating two screens; resync never flushes it.

**6. 🟠 First flip before the GET resolves → the new screen never gets its saved view this session** · `useEntityDataView.js:126`
Resync reads `prefsRef.current` while still `null` (bare seed) and the mount `.then` applies the *old* screenId (#3). No effect re-runs when `prefsRef` later populates, so the flipped-to screen ignores its saved prefs until `screenId` changes again.

**7. 🟠 Every debounced view-save flushes the site-wide resolver cache** · `includes/cascade/class-wp-admin-shell-cache.php:153`
`updated_user_meta` on `wp_admin_shell_user_prefs` → `Cache::flush()` → `DELETE FROM wp_options WHERE option_name LIKE '_transient_wp_admin_shell_resolved_%'` — deletes **every user's** cached resolved config. Pre-PR this meta changed only on rare appearance saves; the PR makes it fire on routine column/sort/perPage changes, so one user fiddling columns forces a full six-origin re-resolve for everyone on their next admin load.

**8. 🟡 Baseline computed from `seed()` but state applied from `current` → spurious write-back** · `useEntityDataView.js:95`
Baseline = `stringify(pickDurableView(applySavedView(seed(), saved)))` while state is set from `applySavedView(current, saved)`. When `current` diverges from `seed()` on a durable axis outside `saved` (async `defaultView`, or a concurrent edit), the next persist run POSTs a non-edit.

**9. 🟡 GET-resolves-after-POST overwrites `prefsRef` with the pre-write blob** · `useEntityDataView.js:89/162`
If the debounced POST resolves first and the slower mount GET resolves after, the `.then` clobbers `prefsRef` with the stale snapshot. On-disk is fine, but a later same-session flip rehydrates stale and drops the just-saved view.

**10. 🟡 Uncancelled timer on a `resyncKeys`-only change writes a stale context's view to the shared `screenId` bucket** · `useEntityDataView.js:124`
Resync doesn't cancel the pending timer. Harmless when `screenId` + `postType` change together (default shell), but a shell mapping two post types to one `screenId` would persist the old post type's view to the shared bucket.

**11. 🟡 Test gap — the suite ships the corruption green** · `tests/runtime/dataviews-shared.test.mjs`
The new tests only round-trip the pure JS helpers (clean object-spread), never the PHP `deep_merge`. A PHP fixture (`existing fields=['a','b','c']`, POST `['a']`, assert stored `=== ['a']`) would have caught #1. Per `CLAUDE.md`, that fixture should land *before* the fix.

### Cleanup / altitude

**12. 🔵 The async GET is avoidable — and removing it closes most of the race window** · `useEntityDataView.js:82`
Prefs are already read server-side as the `user` cascade origin and a config object is already inlined via `wp_add_inline_script`. Inlining the `dataViews` slice into `window.wpAdminShell` (read synchronously, like `useDataView` already does) deletes the per-mount round-trip, the first-paint flash, **and** the async window behind #3/#4/#6/#9. Each of the 6 list apps currently GETs the full blob on every mount with no cross-instance cache.

**13. 🔵 Hand-rolled debounce re-implements `@wordpress/compose` `useDebounce`** · `useEntityDataView.js:155`
The `persistTimer` ref + `setTimeout`/`clearTimeout` + unmount-cleanup effect (~25 lines) duplicates `useDebounce`/`useDebouncedCallback`, already used in `src/runtime/regions/Region.js`.

**14. 🔵 `lastPersisted` is a tri-state sentinel coordinated across 3 effects** · `useEntityDataView.js:76`
Correctness hinges on effect *declaration order* (the comments admit it), splitting the baseline reset/establish across two effects — the root of #5/#8. A single `usePersistedView` primitive driven off the inlined snapshot (#12) would collapse the three-refs/four-effects machine.

---

Most important: **#1** stores corrupt data on the second column-hide regardless of any client change — pair the fix with the **#11** PHP fixture that proves it, and strongly consider **#12** (inline the prefs), which neutralizes four of the React races at the root rather than guarding each.

PR #247, opened 2026-06-04
Add vocabulary & naming spec for WP Admin Workspaces rebrand

dabowman:
## Summary

Introduces `docs/vocabulary-spec.md`, the canonical source of truth for the WP Admin Shell → WP Admin Workspaces rebrand. This spec unifies scattered terminology ("shell", "environment", "admin.json") under a single coherent model and provides the mechanical migration sequence and decision framework for the rename.

## Key changes

- **Unified vocabulary model** (§1–2): Establishes three levels — product ("WP Admin Workspaces"), unit ("a workspace"), and internals (kept architecture terms like `engine`, `region`, `app`). Retires "shell", "shell environment", "environment", and "admin.json" as generic terms.

- **Schema shape evolution** (§3): Dissolves the inner `workspace` block by promoting `engine` and `default-screen` to top-level fields and moving `branding`, `notices`, `widgets`, `customizable` under a new `frame` block. Updates cascade merge semantics, deny-list paths, and `customizable` allowlists accordingly.

- **Config file family alignment** (§4): Renames `admin.json` → `workspace.json` and schema files (`admin*.json` → `workspace*.json`), fixing the prior asymmetry between artifact names and schema names.

- **Bundled workspaces & templates** (§5): Renames `shells/` → `workspaces/` and introduces "template" framing for bundled configs, distinguishing "default workspace" (baseline), "example workspaces" (demos), and "templates" (copy-to-start).

- **Comprehensive prefix map** (§6): Establishes the rule that **all symbols use the plugin slug `wp-admin-workspaces`** (plural everywhere, no exceptions). Includes a full public API rename table and user-facing string updates.

- **Landing page litmus test** (§7): Identifies `docs/index.html` as the acceptance test — the page must read cleanly with zero "shell"/"environment" and "workspace" carrying the weight.

- **Preserved architecture vocabulary** (§8): Explicitly lists terms that do NOT change (`runtime`, `kernel`, `engine`, `region`, `app`, `screen`, `menu`, `command`, `mode`, `slot`, `cascade`, `origin`, `chrome` as styling concept, etc.) because they are descriptive, author-facing, and renaming them would reduce clarity.

- **Role-keyed workspace files** (§9): Designs a greenfield feature (sequenced after the rename) where workspace file names encode the audience (`workspace.json`, `administrator.json`, `editor.json`, etc.), turning the cascade's per-role layer into a legible authoring surface. Flags four open questions (role slug vs alias, location, trust tier, origin mapping) to resolve before building.

- **Migration sequence** (§10): Provides a seven-step plan: spec adoption → schema + shape → prefix codemod → config file rename → public copy → bundled rename → role-keyed files (later). Steps 1–6 are mechanical and fully testable; step 7 is feature work.

- **Open decisions tracker** (§11): Lists five decisions to track (plural prefix rule, role slug choice, directory structure, trust tier, origin mapping, repo rename).

## Notable implementation details

- The spec is explicitly marked as "Adopted for the launch rebrand" and is the canonical source of truth — when this file and prose elsewhere disagree, this file wins until other docs are migrated.
- The one-sentence pitch ("A **workspace** is defined by a `workspace.json` file. It's composed of **apps** mounted in **regions**, rendered by an **engine**.") parallels `theme.json` exactly, reinforcing the analogy the design already reaches for.
- The prefix rule is exceptionless (plural everywhere) to avoid per-symbol ambiguity, even though two spots read slightly long (`wp_admin_workspaces_register_workspace()`, `--wp-admin-workspaces--*`).
- The role-keyed workspace files feature is explicitly deferred and flagged as greenfield work with unresolved security implications (trust tier of file-authored role layers).

https://claude.ai/code/session_01HVjfp4AeXLTDimnL5hipcV


wp-admin-workspaces.php
```diff
@@ -46,101 +46,101 @@ function wp_admin_shell_dependencies_met() {
...
+	if ( $current_version >= WP_ADMIN_WORKSPACES_DB_VERSION ) {
 		return;
 	}
 
 	if ( $current_version < 1 ) {
 		// Step 1 — legacy active-config write-copy.
-		if ( get_option( 'wp_admin_shell_active_shell', '' ) === '' ) {
-			$legacy = get_option( 'wp_admin_shell_active_config', '' );
+		if ( get_option( 'wp_admin_workspaces_active_workspace', '' ) === '' ) {
+			$legacy = get_option( 'wp_admin_workspaces_active_config', '' );
```
dabowman: **This migration step is now dead code — the codemod renamed a stored-data key.**

`wp_admin_workspaces_active_config` is never written anywhere (grep confirms: the only `get_option('wp_admin_workspaces_active_config')` in the tree is this line; every other `wp_admin_workspaces_get_active_config` hit is the *function*, not the option). On `main` this read `wp_admin_shell_active_config` — the genuine MVP legacy key (see `CHANGELOG.md:119,143`). The rename mechanically rewrote the **source** key too, but stored option names on disk are not subject to a code rename, so Step 1 can never find a value: it's a silent no-op, and the `db_version` block around it just stamps and exits.

Given `CHANGELOG.md:5` ("no installed base"), there's no production data to lose — but this still ships a migration that pretends to bridge data and can't. Either:
- point the source read at the real legacy key(s) — `wp_admin_shell_active_config` (and/or the canonical `wp_admin_shell_active_shell`) → `wp_admin_workspaces_active_workspace` — to actually bridge any dev/staging state, **or**
- drop the migration block entirely if no bridge is intended.

As written it's misleading.

---

uninstall.php
```diff
@@ -41,32 +41,32 @@ function wp_admin_shell_uninstall_cleanup() {
 	$wpdb->query(
 		$wpdb->prepare(
 			"DELETE FROM {$wpdb->options} WHERE option_name LIKE %s",
-			$wpdb->esc_like( 'wp_admin_shell_' ) . '%'
+			$wpdb->esc_like( 'wp_admin_workspaces_' ) . '%'
```
dabowman, line 44: **Companion to the migration gap (low severity given no installed base).** This sweep only deletes `wp_admin_workspaces_%`. Because none of the old `wp_admin_shell_*` options/transients/usermeta are migrated or renamed-in-place, an install that had any pre-rebrand state would leave those rows orphaned forever on uninstall — the clean-uninstall guarantee in this file's own docblock doesn't hold for the upgrade path. Moot for fresh installs; only worth fixing if you decide to bridge old→new at all (see the migration comment). If you do bridge by rename-in-place, this becomes a non-issue automatically.

---

dabowman (commented):
## Code review — `wp-admin-shell` → `wp-admin-workspaces` rebrand (max-effort recall pass)

Reviewed the full 448-file diff across nine finder angles (cross-language seams, incomplete renames, over-rename/stability, test+schema+shape correctness, plus reuse/simplification/efficiency/altitude), with independent verification of every candidate.

**Headline: this is a remarkably clean mechanical rebrand.** No security holes, no broken references, no fatal-causing rename gaps. Specifically verified clean:

- **No stale residue** in code — zero `WP_Admin_Shell`, `wp_admin_shell`, `wpAdminShell`, `wp-admin-shell/v1`, or `shells/`/`admin*.json` references survive in any `.php`/`.js`/`.mjs`/`.ts` file (docs/changelog excluded). All requires/imports resolve; class names match filenames.
- **Cross-language seams consistent** — `window.wpAdminWorkspaces` + every sub-key, the `wp-admin-workspaces` script/style handle, `#wp-admin-workspaces` mount, REST namespace `wp-admin-workspaces/v1`, the `wp_admin_workspaces_classic` cookie+nonce, and the iframe-bridge postMessage prefix all match producer↔consumer.
- **WP-core identifiers untouched** — capability strings (`manage_options`, etc.), core hooks, core option keys (`home`/`siteurl`/`default_role`/`gmt_offset`/…), `register_setting` core groups, and `Requires Plugins: gutenberg` were not caught by the codemod.
- **i18n** — all 642 text-domain args are `wp-admin-workspaces`, matching the header.
- **Schema** — `$id`s updated, `$schema` pointers → `workspace.json`, Ajv loader maps to new names; `workspace.json` schema correctly defines the new top-level `engine`/`default-screen`/`frame` shape with `additionalProperties:false`, and bundled configs validate.
- **Config-shape evolution (`workspace` block → top-level `engine`/`default-screen` + `frame`)** — the two runtime-consequential keys have fallback readers (`config.engine || config.workspace?.engine` in `kernel.js`/`buildRuntimeConfig.mjs`; same in `resolver.php`). `frame.{branding,notices}` have no by-name runtime consumer in either version, so the move is inert (no regression).
- **Security** — the trust-tier deny-list correctly moved `workspace.engine` → `engine` (`customizable.php:78`), the single-segment glob matches the top-level path, and `run-security-cascade-tests.php` was updated to assert it. No consumer-origin write path to a gated key was opened by the reshape.

### Findings

**1. (Correctness) Migration Step 1 is dead code** — inline on `wp-admin-workspaces.php:94`. The codemod rewrote the legacy *source* option key, so the v0→v1 migration can never fire. The one genuine defect in the diff.

**2. (Low) `uninstall.php` orphans old-prefix rows** — inline on `uninstall.php:44`. Same root cause; moot for fresh installs.

**3. (Cleanup) `get_branding()` reads a stale path** — `includes/class-wp-admin-workspaces-config.php:95-99` reads `styles.branding` then top-level `branding`, never `frame.branding` where `workspaces/wp-admin-default.json` now puts it. It has **zero production callers** (branding reaches the runtime via engine `defaultRegions` region synthesis, not this getter), so no runtime break — but it's now dead/misleading. Update to `frame.branding` or delete.

**4. (Cleanup) Stale `workspace` blocks in PHP test fixtures** — `run-alpha-trigger-tests.php`, `run-cap-gating-smoke.php`, and `run-appearance-menu-tests.php` still build fixture docs carrying the dissolved `'workspace' => ['engine'=>…,'default-screen'=>…]` block. Tests still pass (none assert on the moved keys; the appearance-menu fixture also sets the new top-level `engine` that production reads), but the fixtures imply the old shape is still meaningful. Reshape them to the new top-level form.

**5. (Cosmetic) Stale doc comments** — `customizable.php:31,66` still say `workspace.engine` though the code uses `engine`; `run-shape-tests.php:11,16` docblock says "`workspace` block present" though it asserts the new top-level shape; `tests/schema/validate-workspaces.test.mjs:28` keeps the var name `SHELLS_DIR` (resolves to `workspaces/` fine) and line 106 references `admin-app.json`/`admin-engine.json`.

### Cross-cutting note
Findings 1, 2, and the un-bridged option/usermeta/override-file renames (`wp-content/admin.json` → `workspace.json` with no fallback) all share one theme: **no old→new storage bridge exists.** Per `CHANGELOG.md:5` ("no installed base") this is safe for users, but it does silently reset any state in your own dev/staging DBs and any `wp-content/admin.json` you've authored. Worth a conscious decision — bridge it, or drop the migration scaffolding that implies you did.

Nothing here blocks the rebrand's correctness against its own new names; #1 is the only behavioral bug and it's a no-op rather than a crash.

---

wp-admin-workspaces.php
dabowman: Fixed in fda9c0c — went with the bridge rather than dropping the block (the team's own dev/staging sites are the de-facto installed base).

- **Step 1** now reads the real on-disk legacy keys: `wp_admin_shell_active_config` → `wp_admin_shell_active_shell` (no more reading the sweep-mangled `wp_admin_workspaces_active_config` that never existed).
- **New Step 2** (`db_version` bumped to `2`) copies every persisted `wp_admin_shell_*` option forward to its `wp_admin_workspaces_*` name — `active_shell`→`active_workspace`, `workspace_enabled`→`enabled`, plus `settings`/`site_config`/`role_config` and the `user_prefs` usermeta key. Idempotent: only seeds a new key when absent, so it can't clobber a later choice. This also closes the silent re-enable you'd get when an explicitly-`false` enable toggle went missing.

The block is now a real bridge, not a no-op.

---

uninstall.php
dabowman: Fixed in fda9c0c. Since I bridged old→new (rather than rename-in-place), `uninstall.php` now sweeps **both** namespaces: a parallel `wp_admin_shell_%` options + `_transient(_timeout)_wp_admin_shell_%` delete and a `wp_admin_shell_user_prefs` usermeta clear, alongside the existing `wp_admin_workspaces_%` sweep. Upgraded installs no longer orphan pre-rebrand rows. Also added an explicit breaking-change + migration note to `CHANGELOG.md`.

---

wp-admin-workspaces.php
```diff
@@ -98,49 +108,86 @@ function wp_admin_shell_dependencies_met() {
...
+			if ( $sentinel !== $old_val && $sentinel === get_option( $new_key, $sentinel ) ) {
+				update_option( $new_key, $old_val );
+			}
+		}
+
+		// User-prefs meta: rename the meta_key for every user that hasn't
+		// already got the new key (guarded so a partial re-run is safe).
+		global $wpdb;
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery -- one-time keyed migration; no caching applies.
+		$wpdb->query(
```
dabowman, line 176: **Minor (very low severity): the raw usermeta rename leaves the `user_meta` object cache stale.** This `UPDATE` mutates `meta_key` directly in the DB, but `get_user_meta()` caches *all* of a user's meta in one blob under the `user_meta` group keyed by `user_id`. On an install with a persistent object cache (Redis/Memcached) where an affected user's meta was already cached, the cached blob still carries the old key `wp_admin_shell_user_prefs` and lacks the new `wp_admin_workspaces_user_prefs` — so `resolver.php:306/331` and the `/user-prefs` controller read **empty prefs** for that user until something busts the cache (a profile save, `clean_user_cache`, or TTL). It self-heals and only bites upgraded installs with a persistent cache, so this isn't a blocker — but a `clean_user_cache()` over the affected user IDs (or a `wp_cache_flush_group('user_meta')`) after the `UPDATE` would close it.

Nit: the Step 2 docblock says "the legacy `wp_admin_shell_*` rows are left in place" — true for the options (copied), but this usermeta path *renames* the key in place, so the old `wp_admin_shell_user_prefs` row no longer exists afterward. Slightly contradicts the comment (and means the matching `delete_metadata` in `uninstall.php` is a no-op for migrated users — harmless).

---

dabowman (commented):
## Re-review — commit `fda9c0c` ("fix: option migration bridge + spec-align two symbols")

Reviewed the incremental change since my last pass. **Both of my findings are resolved, and the fix is sound.**

### ✅ Finding 1 (dead migration step) — FIXED
- DB version bumped `1` → `2`.
- **Step 1** now reads the real on-disk legacy keys (`wp_admin_shell_active_config` → `wp_admin_shell_active_shell`), faithful to the pre-rebrand chain — exactly the correction needed.
- **Step 2** (new) bridges `wp_admin_shell_*` → `wp_admin_workspaces_*` for active-workspace, enable toggle, settings, site-config, and role-config, using a `__wpaw_absent__` sentinel so it copies even falsy values and never clobbers a value the user already set. Idempotent and version-gated. Step 1→Step 2 chain works in a single pass on a v0 upgrade. 👍

### ✅ Finding 2 (uninstall orphans old-prefix rows) — FIXED
`uninstall.php` now sweeps both namespaces — `wp_admin_shell_` options, `_transient(_timeout)_wp_admin_shell_` transients, and the `wp_admin_shell_user_prefs` usermeta — in addition to the new prefix. Clean-uninstall guarantee restored for the upgrade path.

### ✅ Bonus: two spec-aligned public renames — verified complete & consistent
- `wp_admin_workspaces_register_admin_route` → `wp_admin_workspaces_register_route`: function def, the admin-routes docblock, `run-menu-route-shims-tests.php`, CLAUDE.md, and the spec table all updated; no live caller references the old name (only CHANGELOG/archive prose, which is fine).
- `wp_admin_workspaces_workspace_enabled` → `wp_admin_workspaces_enabled`: updated consistently across all 7 sites — `register_setting` (with `show_in_rest` preserved, so the `core:settings-workspace` DataForm still reads/writes it), the classic settings-page form inputs, `wp_admin_workspaces_is_active()`, the settings-workspace app `index.js`, the migration map, the uninstall known-options list, and `run-alpha-trigger-tests.php`. No residual old key in code.

### New finding (very low — inline on `wp-admin-workspaces.php:132`)
The raw `$wpdb` usermeta-key rename doesn't bust the `user_meta` object cache, so on a persistent-cache install an upgraded user's prefs can read empty until the cache is busted. Self-heals; only affects upgraded installs with a persistent cache. Not a blocker; a `clean_user_cache()` after the UPDATE would close it. Plus a one-line docblock nit (the usermeta path renames-in-place, contradicting the "left in place" comment).

### Heads-up (not a code issue)
GitHub reports the branch as `mergeable_state: dirty` — main has advanced and this 449-file rename will need a rebase/merge before it can land.

Net: the migration is now correct and the symbol renames are clean. Nothing here blocks; the one inline note is a polish item.

---

PR #255, opened 2026-06-04
fix: snackbar toasts flush against left viewport edge (#250)

### LINKED ISSUE #250: Snackbar toasts render hard against the left viewport edge
## What happens

Snackbar/toast notifications appear correctly offset from the bottom of the viewport but sit flush against the **left** edge — no left padding/offset.

## Root cause (validated)

`src/apps/notices-snackbar/index.css` pins the list container:

```css
.wp-admin-workspaces-notices-snackbar {
	position: fixed;
	bottom: var(--wpds-dimension-padding-lg);
	right: var(--wpds-dimension-padding-lg);
	…
}
```

…but never sets `width`. The upstream `@wordpress/components` stylesheet ships:

```css
.components-snackbar-list {
	position: absolute;
	z-index: 100000;
	width: 100%;   /* ← survives */
	…
}
```

Both selectors are single-class (equal specificity); the app rule wins on `position`/`bottom`/`right` by load order, but **`width: 100%` is never overridden**. Result: a `position: fixed`, full-viewport-width strip anchored at `bottom`/`right`. Each `.components-snackbar` inside is `width: fit-content` (≥600px media query) inside block `__notice-container` wrappers, so toasts left-align — hard against the viewport's left edge. Bottom offset paints correctly, which matches the report exactly.

## Fix

One-liner in `src/apps/notices-snackbar/index.css`: add `width: fit-content;` (or `width: auto;` + `align-items: flex-end;` on the flex column) so the container shrink-wraps and the existing `right` anchor produces bottom-right stacking as intended.

dabowman:
Adds `width: fit-content` and `align-items: flex-end` to the snackbar container so the existing `right` anchor produces correct bottom-right stacking instead of left-edge flush rendering.

The upstream `@wordpress/components` stylesheet sets `width: 100%` on `.components-snackbar-list`; without an override, the container spans the full viewport width and toasts left-align against the left edge.

Fixes #250

### DIFF
--- a/src/apps/notices-snackbar/index.css
+++ b/src/apps/notices-snackbar/index.css
@@ -5,6 +5,8 @@
 	z-index: 100000;
 	display: flex;
 	flex-direction: column;
+	align-items: flex-end;
+	width: fit-content;
 	gap: var(--wpds-dimension-gap-sm);
 	pointer-events: none;
 }


dabowman:
@claude address feedback

PR #256, opened 2026-06-04
fix: prevent iframe history pollution on Back between iframe screens (#251)

### LINKED ISSUE #251: Browser Back between iframe screens desyncs chrome from content: iframe src swap pollutes session history
## Repro

1. Navigate to **Tools → Export** (`/tools/export`, `iframe:export.php`, default mode).
2. Navigate to **Theme File Editor** (`/tools/theme-editor`, `iframe:theme-editor.php`, `mode: takeover`).
3. Hit browser **Back**.

**Observed:** the Export page content renders, but the chrome stays in **takeover** mode.

## Root cause (validated)

Both screens resolve to the same app source. `iframe:<slug>` refs are rewritten to `core:iframe-fallback` (`src/runtime/compile/translateIframeRef.mjs`), so navigating export → theme-editor renders the **same component type at the same tree position** — `renderRegionApp()` in `src/runtime/regions/Region.js` mounts `<MountedApp appRef={ref} regionId={region.id} />` with **no `key`**, and `MountedApp` adds none either.

React therefore reuses the `IframeApp` instance and the underlying `<iframe>` DOM node, updating only the `src` prop (`src/apps/iframe-fallback/index.js` renders `<iframe src={src} …>`). Per the HTML spec, **setting `src` on an iframe already in the document is a navigation that adds a joint session-history entry** (only the initial insertion replaces).

History after step 2:

1. `#/tools/export` (top-window route entry)
2. `#/tools/theme-editor` (top-window route entry)
3. iframe navigation `export.php` → `theme-editor.php` (joint entry, top URL unchanged)

Back pops entry 3: the **iframe** travels back to `export.php` while the top window stays on `#/tools/theme-editor`. `useMode()` resolves the mode from the URL-matched screen (`src/runtime/modes/useMode.js`), which is still `theme-editor` → takeover persists. Content and chrome are now permanently desynced until another Back/forward.

The mode system itself is correct — it's pure and URL-driven. The bug is history pollution from the reused iframe element.

## Fix options

- **Remount per route**: key the app mount on the matched route/app identity (e.g. `key={matched?.app + JSON.stringify(matched?.config)}` or the screen id) in `renderRegionApp` — a freshly inserted iframe's initial load **replaces** instead of pushing, so no stray entry. Cheapest and also resets per-screen component state that arguably shouldn't survive across screens.
- **Navigate via `location.replace`**: in `IframeApp`, when `src` changes after mount, call `iframeRef.current.contentWindow.location.replace(src)` instead of letting React mutate the attribute.

Note the same trap exists inside `IframeApp` already: the bridge's `onIframeNavigate` (`iframeRef.current.src = href`) and the re-auth path (`iframe.src = iframe.src`) both push history entries; worth auditing in the same pass.

Related: #252 (duplicated file-editor nav entries) is how this repro was found.

dabowman:
Fixes #251

Two complementary fixes for the chrome/content desync when pressing Back between screens that both resolve to `core:iframe-fallback`.

**Region.js:** Key the app-mount wrapper on `source + JSON(config)` so React unmounts and remounts IframeApp when the route identity changes. A freshly inserted iframe's first src load is a replace rather than a push, eliminating the stale joint session-history entry.

**IframeApp:** Use `contentWindow.location.replace()` instead of `iframe.src =` mutation for both `onIframeNavigate` and the re-auth reload path.


dabowman:
@claude address the feedback and failing CI

PR #262, opened 2026-06-04
fix(posts): lock pinned-status screens against status-filter changes (#209)

### LINKED ISSUE #209: Lock pinned-filter list screens (Trash / Drafts / Pending) against status-filter changes
## Problem

The dedicated `core:posts` status screens — **Trash**, **Drafts**, **Pending** — pin a status via both `config.status` and the variant's `defaultView.filters`, but the inherited `status` field stays **user-changeable** in the DataViews UI.

If a user switches the status filter while on one of these screens, `queryArgs.status` follows it (`src/apps/posts/index.js`), so the list fetches posts of a different status than the screen's label implies. The most visible case is **Trash**: switching the filter to `publish` fetches non-trashed posts, and because `edit`/`view`/`trash` are gated by `eligibleWhen: { status: [publish, …] }`, those actions reappear on rows inside a screen labelled "Trash". Classic wp-admin locks each of these views to its status.

## Scope

This is a pre-existing pattern, not introduced by #195 — the `drafts` and `pending` variants have the same pinned-but-changeable filter. It was surfaced as an out-of-scope, non-blocking nit during the review of PR #195 (parity item 8 / issue #104):
- https://github.com/dabowman/WordPress-Admin-Environment/pull/195#discussion_r3329176915

A proper fix should cover **all** pinned-filter screens uniformly, not bolt a one-off onto `trash`.

## Possible approaches

- Make the `status` field **non-filterable** in these variants (drop `filterBy` / mark not-filterable) so the pinned status can't be overridden from the UI — while keeping it as a display/sort column.
- Or hard-pin the status in the query regardless of the active filter when the screen declares `config.status` (ignore a status filter that contradicts the pin).

Either way, the behavior should be consistent across `trash` / `drafts` / `pending` (and any future pinned-status variant), and ideally generic enough that other entity-CRUD list apps (comments spam/trash, etc.) can opt in.

## Acceptance

- On the Trash / Drafts / Pending screens, the user cannot change the list away from the screen's pinned status via the DataViews filter UI.
- Gated row actions (`edit`/`view`/`trash` on Trash) never reappear on those screens.
- Consistent across the affected screens; no regression to the All Posts screen's freely-filterable status.

Source: parity follow-up from #104 / PR #195. Detail in `docs/parity/posts.md`.

dabowman:
Closes #209.

Trash / Drafts / Pending screens pinned a status but left the status field user-changeable, so switching the filter fetched the wrong status and revived gated row actions. Fixed generically via a new pure helper `_shared/dataviews/pinnedStatus.mjs`: hard-pin the REST query, strip the status field's `filterBy`, and hide the status tab strip when `config.status` names a concrete status. `any`/absent (All Posts) stays freely filterable. Adds helper unit tests + docs.


dabowman:
@claude address the rest of the issues that review found.

